PR #1, opened 2021-04-13
DE-2031: Fix silently ignoring invalid timezone

robjhornby-ts:
Fixed bug that an invalid timezone like "EDT" (which isn't in the [tz timezone database](https://en.wikipedia.org/wiki/List_of_tz_database_time_zones)) would silently fail and default to using no timezone.

Also:
- Simplified timezone parsing logic - we were repeating parts of what arrow.get already does (parsing timezone strings into datetime.tzinfo objects with error handling).
- Fixed bug in test cases - when asserting that errors were raised using try/catch, if the code didn't raise an error, the test would pass incorrectly. Now using pytest.raises to assert errors.
- Refactored tests to use pytest parametrize so that it's clearer which cases pass or fail (was useful for debugging)

### DIFF
--- a/__tests__/unit/test_convert_datetime_to_ts_format.py
+++ b/__tests__/unit/test_convert_datetime_to_ts_format.py
@@ -1,28 +1,34 @@
-import sys
+import pytest
 from dateutil import tz
 from task_script_utils.convert_datetime_to_ts_format import convert_datetime_to_ts_format
 
 
-def test_no_timezone():
-    raw_expect = [
+@pytest.mark.parametrize(
+    "raw,expect",
+    [
         ("2019-07-17 11:21:00", "2019-07-17T11:21:00.000"),
         ("2020-06-18 13:17:57.54036", "2020-06-18T13:17:57.540"),
     ]
-    for raw, expect in raw_expect:
-        assert convert_datetime_to_ts_format(raw) == expect
+)
+def test_no_timezone(raw, expect):
+    assert convert_datetime_to_ts_format(raw) == expect
 
 
-def test_embedded_timezone():
-    raw_expect = [
+@pytest.mark.parametrize(
+    "raw,expect",
+    [
         ("2020-03-06T17:19:45.706000-05:00", "2020-03-06T22:19:45.706Z"),
         ("2020-04-30T20:27:41.000Z", "2020-04-30T20:27:41.000Z"),
+        ("2020-04-30T10:27:41.000 EDT", "2020-04-30T14:27:41.000Z"),
     ]
-    for raw, expect in raw_expect:
-        assert convert_datetime_to_ts_format(raw) == expect
+)
+def test_embedded_timezone(raw, expect):
+    assert convert_datetime_to_ts_format(raw) == expect
 
 
-def test_input_timezone():
-    raw_timezone_expect = [
+@pytest.mark.parametrize(
+    "raw,timezone,expect",
+    [
         ("2019-07-17 11:21:00", "GMT-5", "2019-07-17T16:21:00.000Z"),
         (
             "2019-07-17 11:21:00",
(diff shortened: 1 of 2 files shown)


task_script_utils/convert_datetime_to_ts_format.py
```diff
@@ -30,29 +40,20 @@ def convert_datetime_to_ts_format(
...
-        utc_indicator = "Z"
-
-    else:
-        datetime_iso_local = arrow.get(str(parsed_time), tzinfo=tz.gettz("GMT"))
+    datetime_iso_local = arrow.get(
+        str(parsed_time),
+        tzinfo=timezone_to_use
+    )
 
     ts_datetime = (
```
tlitetrasci, line 58: Can simplify to:

```
return (datetime_iso_local
    .to()
    etc
)
```

No need to create extra variables

__tests__/unit/test_convert_datetime_to_ts_format.py
```diff
@@ -1,28 +1,34 @@
-import sys
+import pytest
 from dateutil import tz
 from task_script_utils.convert_datetime_to_ts_format import convert_datetime_to_ts_format
 
 
-def test_no_timezone():
-    raw_expect = [
```
tlitetrasci, line 7: I like the conversion to parameterized tests.

I'm wondering how stringent we want this function to be. @wkopen What's using this at the moment? Personally I tend to have unique time parsing code for each instrument, since it constrains the problem.

Here are some example tests I wrote for the BMG date parser:
```


@pytest.mark.parametrize("tstamp, day_first", [
    # test swapping day/month order
    ("01.05.2020 07:57:30 (UTC+2)", True),
    ("01.05.2020 07:57:30 (UTC+2)", False),

    # test no time zone
    ("01.05.2020 07:57:30", True),

    # test using dashes as date separators
    ("03-09-2018 17:18:48", True),

    # test AM/PM parsing
    ("8/24/2020 5:53:52 PM (UTC+2)", False),
    ("8/24/2020 5:53:52 pm", False),

    ("4/30/2020 11:40:46 AM (UTC+2)", False),
    ("4/30/2020 11:40:46 am (UTC+2)", False),

    # test no leading zeroes for month
    ("1.5.2020 07:57:30", True),
    ("01.5.2020 07:57:30", True),

    # test no leading zeroes for day
    ("1.5.2020 07:05:30", False),
    ("1.05.2020 07:05:30", False),
])
def test_parse_good_timestamps(snapshot, tstamp: str, day_first: bool):
    res = parse_timestamp(tstamp, day_first)

    snapshot.assert_match(res)


@pytest.mark.parametrize("tstamp, day_first", [
    # test bad month
    ("8/24/2020 5:53:52 PM (UTC+2)", True),

    # test more than 24 hours
    ("4/30/2020 25:40:46 (UTC+2)", False),

    # test mixing separators
    ("8.24/2020 5:53:52 PM (UTC+2)", False),

    # test missing hour separator
    ("8/24/2020 553:52 PM (UTC+2)", False),

    # test time zone formatting
    ("8/24/2020 5:53:52 PM UTC+2)", False),
    ("8/24/2020 5:53:52 PM UTC+2", False),

    # test minutes not using leading zero
    ("8/24/2020 5:2:52 PM (UTC+2)", False),

    # test seconds not using leading zero
    ("8/24/2020 5:02:2 PM (UTC+2)", False),

    # test weird AM/PM formatting
    ("8/24/2020 5:53:52 pM (UTC+2)", False),
    ("8/24/2020 5:53:52 Am (UTC+2)", False),
])
def test_parse_bad_timestamps(tstamp: str, day_first: bool):
    with pytest.raises(Exception):
        res = parse_timestamp(tstamp, day_first)

```

The tests are really trying to probe the boundaries of what we expect to see in the raw files. Since this is such a general time parsing function, I wonder what the boundaries of this function are.

tlitetrasci (changes requested):
What's the difference between using `arrow` and `dateparser`? Why do we need both?

robjhornby-ts:
> What's the difference between using `arrow` and `dateparser`? Why do we need both?

I suspect the reason dateparser was used is because it handles non-ISO datetime strings by guessing the datetime format, which arrow doesn't do.

For example:

```
print(dateparser.parse("2020-04-30T10:27:41.000 EDT"))
```
output: `2020-04-30 10:27:41-04:00`

```
print(arrow.get("2020-04-30T10:27:41.000 EDT"))
```
raises `ParserError: Expected an ISO 8601-like string, but was given '2020-04-30T10:27:41.000 EDT'. Try passing in a format string to resolve this.`


I would be in favour of removing `dateparser` because we generally shouldn't trust the automatic format parsing with all the possible ambiguities it introduces (e.g. day-first-date format - `dateparser` has no way of knowing which to use). We know enough about the input files that we should always be able to determine the datetime format, which removes this potential source of bugs. On the other hand, several of the pipelines currently using this function rely on the `dateparser` path of the logic so this would not be a backward-compatible change. @wkopen what do you think?

task_script_utils/convert_datetime_to_ts_format.py
```diff
@@ -1,24 +1,34 @@
...
+            If timezone is defined, the output is in UTC, indicated by 'Z'.
+            Otherwise, the output will be timezone unaware, indicated by 
+            having no 'Z' at the end.
+
+        Examples:
+        If timezone is defined: "2019-07-17T14:21:00.000Z"
+        If timezone is not defined: "2019-07-17T11:21:00.000"
     """
 
     if datetime_format:
```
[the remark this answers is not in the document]
robjhornby-ts: This function currently relies on the library functions it calls to raise any errors, so it's just acting as a wrapper to those functions. I agree it could be better to validate the inputs to this this function so we have more control over how it's used - it seems like there are design questions which need answering which are bigger than this bugfix.

task_script_utils/convert_datetime_to_ts_format.py
```diff
@@ -30,29 +40,20 @@ def convert_datetime_to_ts_format(
 
     if parsed_time is None:
         raise ValueError(
-            f"Could not parse input datetime string {datetime} using {parser_name} and the following input formats: {datetime_format}"
+            f"Could not parse input datetime string {datetime} using {parser_name} and the following input formats: '{datetime_format}'"
         )
```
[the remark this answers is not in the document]
robjhornby-ts: That's how I interpreted it but actually I think we should limit it to one datetime format and enforce that, I can't think of any use case for multiple formats - I'll make that change.

task_script_utils/convert_datetime_to_ts_format.py
```diff
@@ -30,29 +40,20 @@ def convert_datetime_to_ts_format(
...
-            )
-            timezone_to_use = timezone
+    utc_indicator = "Z"
+    timezone_to_use = timezone if timezone else parsed_time.tzinfo
 
-    datetime_iso_local = ""
-    utc_indicator = ""
+    if not timezone_to_use:
+        timezone_to_use = tz.gettz("GMT")
+        utc_indicator = ""
```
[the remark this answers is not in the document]
robjhornby-ts: Agreed, I'll tidy that up

task_script_utils/convert_datetime_to_ts_format.py
```diff
@@ -1,24 +1,34 @@
...
+
+    Output:
+        Datetime string in ISO-8601 format with millisecond precision.
+            If timezone is defined, the output is in UTC, indicated by 'Z'.
+            Otherwise, the output will be timezone unaware, indicated by 
+            having no 'Z' at the end.
+
+        Examples:
+        If timezone is defined: "2019-07-17T14:21:00.000Z"
+        If timezone is not defined: "2019-07-17T11:21:00.000"
```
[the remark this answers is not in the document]
robjhornby-ts: I'll make the examples more complete

task_script_utils/convert_datetime_to_ts_format.py
```diff
@@ -30,29 +40,20 @@ def convert_datetime_to_ts_format(
...
-
-    else:
-        datetime_iso_local = arrow.get(str(parsed_time), tzinfo=tz.gettz("GMT"))
+    datetime_iso_local = arrow.get(
+        str(parsed_time),
+        tzinfo=timezone_to_use
+    )
 
     ts_datetime = (
         datetime_iso_local.to("GMT")
```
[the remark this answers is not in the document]
robjhornby-ts: Changed

task_script_utils/convert_datetime_to_ts_format.py
robjhornby-ts: Agreed

__tests__/unit/test_convert_datetime_to_ts_format.py
robjhornby-ts: > What's using this at the moment?

I searched `ts-lib-protocol-script` for `task_script_utils.convert_datetime_to_ts_format` to get an idea of which task scripts are using this version of the function (this might not catch all uses).
To summarize, here are the arguments currently being used with this function:

`datetime` only: 10
`datetime` and `datetime_format`: 1
`datetime` and `timezone`: 1
`datetime`, `datetime_format` and `timezone`: 0

So `datetime` only is the most common input - handled by `dateparser` since there's no format string given. 

> I wonder what the boundaries of this function are

It seems like there are two parts to the functionality:

1. Convert parsed datetimes to the TS standard string format
2. Parse datetime strings

As a minimum, I think we should always use a function which implements part 1. This is common to all parsers which deal with datetimes, and it provides a single source of truth for the format which goes into an IDS - e.g. if we ever needed to change how datetimes are represented, we would only need to change it it one place.

I think it also makes sense to have part 2 be in a common library - datetime parsing is a general problem so it makes sense to have a general function shared across all parsers. This means all testing, maintenance and bugfixes only have to happen in one place instead of spread out and repeated across all parsers - your tests are all general for datetime parsing rather than instrument-specific so could be added to this function's tests.

This function is currently implementing these two parts. Then there's the third part which can't go in a central place:

3. Instrument-specific datetime parsing

This should only be needed if the instrument is using an unconventional datetime output, which is likely to be rare. This part has to live in each separate task script and can still make use of central functions once the instrument-specific part is parsed, to give a consistent output.


Do you think we should decide on all of these points before this bugfix is merged? Any bigger change in how this function works could also indirectly solve this bug, and potentially other unfound bugs, so that could make sense. @wkopen what do you think?

wkopen:
> > What's the difference between using `arrow` and `dateparser`? Why do we need both?
> 
> I suspect the reason dateparser was used is because it handles non-ISO datetime strings by guessing the datetime format, which arrow doesn't do.
> 
> For example:
> 
> ```
> dateparser.parse("2020-04-30T10:27:41.000 EDT")
> ```
> 
> output: `datetime.datetime(2020, 4, 30, 10, 27, 41, tzinfo=<StaticTzInfo 'EDT'>)`
> 
> ```
> arrow.get("2020-04-30T10:27:41.000 EDT")
> ```
> 
> raises `ParserError: Expected an ISO 8601-like string, but was given '2020-04-30T10:27:41.000 EDT'. Try passing in a format string to resolve this.`
> 
> I'm conceptually in favour of removing `dateparser` because we generally shouldn't trust the automatic format parsing with all the possible ambiguities it introduces (e.g. day-first-date format - `dateparser` has no way of knowing which to use). We know enough about the input files that we should always be able to determine the datetime format, which removes this potential source of bugs. On the other hand, several of the pipelines currently using this function rely on the `dateparser` path of the logic so this would not be a backward-compatible change. @wkopen what do you think?

I agree with this as we are moving towards a more precise data processing strategy (more permissive doesn't mean we will forgo quality). And using `dateparser` because it can try to guess the date and time is the opposite of what we want to do. If we make the change we need to bump the major version of this util library.

__tests__/unit/test_convert_datetime_to_ts_format.py
wkopen: I agree with Rob that we need a shared util code to handle time conversion. @tlitetrasci I'm curious why your test cases are specific to certain instruments. As Rob said, unless the instrument is using something unconventional, we should be able to define the date-time pattern for it and use this util function.

Toby, the test cases you provided above to test the boundary, to me it's more like testing the boundary of `arrow` or this util function, right?

And this topic shouldn't block this PR from merging because this PR fixes a bug in the existing function. If we need to further discuss it, we can start with a meeting. But it should not block this PR.

task_script_utils/convert_datetime_to_ts_format.py
```diff
@@ -1,24 +1,34 @@
...
-            If timezone is not defined: "2019-07-17T11:21:00.000"
+
+    Inputs:
+        datetime - datetime string
+        datetime_format - datetime format string
+            (must follow https://arrow.readthedocs.io/en/stable/#format)
+        timezone - user-defined timezone. If specified, this overwrites any
+            timezone in the input datetime. It can be a string or a timezone
+            type accepted by arrow.get. Timezone strings must come from the tz
+            timezone database, e.g. "EDT" is not allowed, but is equivalent to
```
wkopen, line 20: Is `EDT` really not allowed or it will not work if provided?

__tests__/unit/test_convert_datetime_to_ts_format.py
```diff
@@ -31,53 +37,94 @@ def test_input_timezone():
         ),
         ("2019-07-17 11:21:00", tz.tzutc(), "2019-07-17T11:21:00.000Z"),
         ("2020-04-30T20:27:41.000Z", "GMT+3", "2020-04-30T17:27:41.000Z"),
+        ("2020-04-30T10:27:41.000", "America/New York", "2020-04-30T14:27:41.000Z"),
```
wkopen, line 40: Can you add another one which is when US is in daylight saving so the difference with UTC+0 is actually 5 hours? Like 2020 Jan.1st

wkopen (changes requested):
Looks good to me overall. Just some minor things. And I think the more general design discussion is out-of-scope of this ticket and shouldn't prevent it from merging. We can discuss that offline.

PR #2, opened 2021-10-11
DE-2591 datetime parser

setiat:


mlipschultz (changes requested):
I feel like I left a lot of comments, but overall it looks like you captured the flow we talked about and implemented the parsing rules.  Good job!

Pipfile
```diff
@@ -13,6 +13,7 @@ numpy = "*"
 python-dateutil = "*"
 dateparser = "*"
 arrow = "*"
+ipykernel= "*"
```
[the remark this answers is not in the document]
setiat: This Pipfile was not updated. latest one has been pushed to repo
- ipykernel is moved to dev requirements, I use it as an alternative to python shell for quick experimenting
- dev packages also contain `autopep8`

task_script_utils/datetime_parser/parser.py
```diff
@@ -0,0 +1,99 @@
...
+        return parsed_datetime
+
+    # Otherwise use DateInfo Parser to parse short dates
+    try:
+        dt_info = DateTimeInfo(datetime_str, config)
+        dt_info.parse()
+        if dt_info.dtstamp:
+            parsed_datetime = dt_info.datetime
+    except Exception as e:
+        raise(e)
```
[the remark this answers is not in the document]
setiat: removed.

task_script_utils/datetime_parser/datetime_info.py
```diff
@@ -0,0 +1,562 @@
...
+        day, month = self._process_day_and_month(others)
+        year = int(year)
+        return day, month, year
+
+    def _process_two_digit_date_pattern(self, matches):
+        if len(matches) > 1:
+            raise MultipleDatesError(f"Multiple Dates Detected: {matches}")
+
+        date = re.sub(r"[-~!@#$%^&*.,;/\\]", "-", matches[0])
+        date_tokens = [i for i in date.split("-")]
```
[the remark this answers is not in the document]
setiat: my bad! Fixed :)

task_script_utils/datetime_parser/datetime_info.py
```diff
@@ -0,0 +1,562 @@
...
+            # 5:30 --> 05:30
+            # 05:30 --> 05:30
+            if len(offset) == 4:
+                return f"0{offset}"
+            elif len(offset) == 5:
+                return offset
+            else:
+                return None
+        else:
+            # 2 --> 02:00 | 09 --> +09:00 | 12 --> 12:00
```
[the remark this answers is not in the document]
setiat: fixed

task_script_utils/datetime_parser/datetime_info.py
```diff
@@ -0,0 +1,562 @@
...
+            and self.hour
+            and self.minutes
+            and self.seconds
+        ):
+            dt_str = f"{self.day}-{self.month}-{self.year}"
+            dt_str += f" {self.hour}:{self.minutes}:{self.seconds}"
+            if self.milliseconds:
+                dt_str += f".{self.milliseconds}"
+
+            if self.am_or_pm and int(self.hour) <= 12:
```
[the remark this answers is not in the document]
setiat: In this case "2018-13-09T16:15+2 AM" will be parsed to "13-09-2018 16:15:00 +02:00"
As long as date_time string is not entered by a human manually, I don't this case will ever occur.
or if it occur, it will be like "13:00 PM", and this is not harmful
But if it not the case, parser should return None.

This raises one more point that I assumed.... If AM/PM is not detected , I am treating time_string  as military time//24hrs format.

task_script_utils/datetime_parser/datetime_info.py
```diff
@@ -0,0 +1,562 @@
...
+        date = re.sub(r"[-~!@#$%^&*.,;/\\]", "-", matches[0])
+        date_tokens = [i for i in date.split("-")]
+
+        if self.config.year_first is True:
+            if self.config.day_first is True:
+                # Input = YY-DD-MM
+                year, day, month = date_tokens
+            elif self.config.day_first is False:
+                # Input = YY-MM-DD
+                # TODO: How to handle 01-15-11
```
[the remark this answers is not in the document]
setiat: cc @robjhornby-ts 
original Conversation : https://tetrascience.atlassian.net/browse/DE-2591?focusedCommentId=49636

task_script_utils/datetime_parser/datetime_info.py
```diff
@@ -0,0 +1,608 @@
...
+            for var_name, func in checks.items():
+                if getattr(self, var_name) is None:
+                    result = func(token)
+                    if result:
+                        setattr(self, var_name, result)
+
+        # if self.iana_tz:
+        #    self.offset_ = pendulum.now(tz=self.iana_tz).format("Z")
+
+    def get_checks(self):
```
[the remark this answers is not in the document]
setiat: These are now protected methods

task_script_utils/datetime_parser/datetime_info.py
```diff
@@ -0,0 +1,608 @@
...
+        self.config = config
+
+    def __str__(self):
+        return json.dumps(self.__dict__, indent=2)
+
+    def parse(self):
+        checks = self.get_checks()
+        tokens = self.date_time_raw.split()
+        for token in tokens:
+            t = token
```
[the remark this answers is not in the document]
setiat: removed

task_script_utils/datetime_parser/datetime_info.py
```diff
@@ -0,0 +1,608 @@
...
+        year_last_pattern = r"\d{1,2}[-~!@#$%^&*.,;/\\]\d{1,2}[-!~@#$%^&*.,;/\\]\d{4,4}"
+        two_digit_date_pattern = r"\d{1,2}[-~!@#$%^&*.,;/\\]\d{1,2}[-!~@#$%^&*.,;/\\]\d{1,2}"
+        # no_sep_date_pattern = r"\d{6,6}"
+
+        # YYYY-XX-XX
+        year_first_matches = re.findall(year_first_pattern, token)
+        if year_first_matches:
+            day, month, year = self._process_year_first_or_last_matches(
+                year_first_matches, True)
+            self.day, self.month, self.year = day, month, year
```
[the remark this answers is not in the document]
setiat: fixed.
No matcher methods will alter instance variable.
It will just return the result

task_script_utils/datetime_parser/datetime_info.py
```diff
@@ -0,0 +1,608 @@
...
+
+        Returns:
+            tuple: If date is parsed successfully return
+            (True, datetime object) else return (False, None)
+        """
+        try:
+            parsed = pendulum.from_format(date_str, format)
+            return True, parsed
+        except Exception as e:
+            return False, None
```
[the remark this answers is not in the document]
setiat: fixed

task_script_utils/datetime_parser/parser.py
```diff
@@ -0,0 +1,149 @@
...
+from .datetime_info import DateTimeInfo
+from .date_formats import get_long_datetime_formats
+from .utils import (
+    convert_offset_to_seconds,
+    replace_abbreviated_tz_with_utc_offset,
+    replace_zz_with_Z
+)
+
+
+default_pipline_config = PipelineConfig(**{})
```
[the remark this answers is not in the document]
setiat: added `DEFAULT_PIPELINE_CONFIG` in `PipelineConfig.py`. Other modules will just import it.

task_script_utils/datetime_parser/parser.py
```diff
@@ -0,0 +1,149 @@
...
+
+
+def parse(
+    datetime_str: str,
+    formats_list: list = None,
```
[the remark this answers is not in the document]
setiat: I changed it to `fomats_list: List[str] = []` 
Let me know if it works.

task_script_utils/datetime_parser/parser.py
```diff
@@ -0,0 +1,149 @@
...
+
+    # Parse Using dateutil.parser.parse
+    parsed_datetime = parse_using_dateutils(datetime_str, config)
+    if parsed_datetime:
+        parsed_datetime = change_fold(parsed_datetime, config.fold)
+        return parsed_datetime
+
+    # Otherwise use DateInfo Parser to parse short dates
+    dt_info = DateTimeInfo(datetime_str, config)
+    dt_info.parse()
```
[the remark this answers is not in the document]
setiat: `DateTimeInfo.parse()` was just place holder. I was planning to add two more function: `parser_short_dates` and `parse_long_dates` and give `Parser.py` the power to decide which fx to call and when.

After implementing these methods, it make more sense to make them protected

task_script_utils/datetime_parser/convert_to_iso_format.py
```diff
@@ -0,0 +1,36 @@
+from typing import List
+
+import pendulum
+from pendulum.datetime import DateTime as PendulumDateTime
+
+from .pipeline_config import PipelineConfig
+from .parser import parse
+
+
+def convert_to_iso8601(
```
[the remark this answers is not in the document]
setiat: added return type and updated doc strings

task_script_utils/datetime_parser/convert_to_iso_format.py
```diff
@@ -0,0 +1,36 @@
...
+    datetime_str: str,
+    formats_list: List[str] = None,
```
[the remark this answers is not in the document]
setiat: same as previous comment: `formats_list: List[str]=[]`

task_script_utils/datetime_parser/parser.py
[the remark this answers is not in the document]
setiat: So now we have.
`_parse()`
`_parse_short_date_formats()`: called by `__init__`
`_parse_long_date_formats()`: called by `long_date_format`

task_script_utils/datetime_parser/datetime_info.py
```diff
@@ -0,0 +1,777 @@
...
+                return f"{offset[:2]}:{offset[2:]}"
+            else:
+                raise InvalidOffsetError(offset)
+
+    def _replace_single_characters(self, string: str):
+        # Input =  2018-13-09T11:12:23.000-05:30
+        # output = 2018-13-09 11:12:23.000-05:30
+        prev_char: str = ""
+        indexes = []
+        for i, current_char in enumerate(string[:-1]):
```
abclay, line 707: This function might be simpler if you:

1. Change the string to a list of characters: `chars = list(string)`
2. Loop through the list, changing whatever you need to change.  Not great example: `chars[5] = " "`
3. Create the final string: `return "".join(chars)`

You can also skip keeping track of prev, cur, and next characters by just starting from the 2nd element. Something like this: 

```python
char_list = list(string)
for i, char in enumerate(char_list[1:-1]):
    if (
        char_list[i-1].isdigit()
        and char_list[i].isalpha()
        and char_list[i+1].isdigit()
    ):
        char_list[i] = " "
return "".join(char_list)
```

abclay (commented):
It would be useful to have a README file for the functionality offered by this library.  Especially for someone new like me, who has not been in any of the design decisions.

__tests__/unit/test_datetime_info.py
```diff
@@ -0,0 +1,130 @@
...
+    ("2021/11/07", None, None, None),
+    ("2021/32/07", None, True, None),
+    ("2021/11/14", None, True, None),
+]
+
+# Test them with year_first
+# The goal is to test regex
+regex_test_cases = [
+    ("1:2:32 20-12-1 AM America/Chicago", "01-12-2020 1:2:32 AM America/Chicago"),
+    ("1:2:32 20-11-1 -1 AM America/Chicago", "01-11-2020 1:2:32 AM -01:00 America/Chicago"),
```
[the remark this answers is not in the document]
setiat: These test cases were written only to check if regex parsing and validation works or not.
`convert_to_ts_iso_8601` takes care of the final formatting.

task_script_utils/datetime_parser/datetime_info.py
setiat: fixed.

task_script_utils/datetime_parser/datetime_info.py
```diff
@@ -0,0 +1,777 @@
...
+            fractional_seconds is not None
+            and len(fractional_seconds) > 6
+        ):
+            raise InvalidDateError(
+                f"Invalid time : {time_}. Fractional Seconds value is incorrect"
+            )
+
+        if not (0 <= int(hour) <= 24):
+            raise InvalidTimeError(
+                f"Invalid time : {time_}. Hours value is incorrect"
```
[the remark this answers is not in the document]
setiat: Done.

task_script_utils/datetime_parser/datetime_info.py
```diff
@@ -0,0 +1,777 @@
...
+    def _parse_short_date_formats(self):
+        short_date_matcher = self._get_matchers_map()
+        self._parse(short_date_matcher)
+
+    def _get_matchers_map(self, long_date_formats=False):
+        """It creates a `dict` that maps parsing functions to instance
+        variable that should store the result.
+
+        eg. if `self.match_short_date` is able to parse short
+        date successfully, it will be stored in `self.short_date`
```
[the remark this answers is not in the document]
setiat: updated doc strings

task_script_utils/datetime_parser/datetime_info.py
```diff
@@ -0,0 +1,777 @@
...
+                if getattr(self, var_name) is None:
+                    result = func(token)
+                    if result and isinstance(result, dict):
+                        for key, value in result.items():
+                            setattr(self, key, value)
+                    elif result is not None:
+                        setattr(self, var_name, result)
+
+        # if self.iana_tz:
+        #    self.offset_ = pendulum.now(tz=self.iana_tz).format("Z")
```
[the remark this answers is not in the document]
setiat: removed.

task_script_utils/datetime_parser/datetime_info.py
```diff
@@ -0,0 +1,777 @@
...
+
+from .parser_exceptions import *
+from .pipeline_config import PipelineConfig
+from .utils import get_time_formats_for_long_date
+
+
+class DateTimeInfo:
+    def __init__(self, date_time_raw: str, config: PipelineConfig):
+        self.date_time_raw = date_time_raw
+        self._pre_process_datetime_string()
```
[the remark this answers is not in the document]
setiat: I agree. made the change

task_script_utils/datetime_parser/convert_to_iso_format.py
```diff
@@ -0,0 +1,43 @@
...
+import pendulum
+from pendulum.datetime import DateTime as PendulumDateTime
+
+from .pipeline_config import PipelineConfig, DEFAULT_PIPELINE_CONFIG
+from .parser import parse
+
+
+def convert_to_ts_iso8601(
+    datetime_str: str,
+    formats_list: List[str] = [],
```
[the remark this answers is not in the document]
setiat: it is now `formats_list: Tuple[str] = (), `

__tests__/unit/test_datetime_parser.py
```diff
@@ -0,0 +1,140 @@
...
+format_list_test_cases = {
+    "Sunday, May 26th 2013 12:12:12 AM IST Asia/Kolkata": "2013-05-26T00:12:12+05:30",
+    "Sunday, May 26th 2013 12:12:12 AM ZST Asia/Kolkata": None,
+    "Sunday, May 26th 2013 12:12:12 AM BST": "2013-05-26T00:12:12+01:00",
+    "Sunday, May 26th 2013 12:12:12 AM Asia/Kolkata": "2013-05-26T00:12:12+05:30",
+    "Sunday, May 26th, 2013 12:12:12 AM Asia/Kolkata": "2013-05-26T00:12:12+05:30",
+    "Sunday, May 26 2013 12:12:12 AM Asia/Kolkata": "2013-05-26T00:12:12+05:30",
+}
+
+prase_with_no_format_list_test_cases = {
```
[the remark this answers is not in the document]
setiat: @abclay Please help,  I am not sure what changed here?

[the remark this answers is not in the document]
setiat: Thanks 🙈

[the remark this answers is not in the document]
setiat: fixed.

__tests__/unit/test_datetime_parser.py
```diff
@@ -0,0 +1,140 @@
...
+
+    if parsed_datetime is None:
+        assert parsed_datetime == expected
+    else:
+        assert parsed_datetime.isoformat() == expected
+
+
+@pytest.mark.parametrize(
+    "input, expected",
+    prase_with_no_format_list_test_cases.items()
```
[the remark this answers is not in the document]
setiat: fixed

abclay (approved):
All my concerns have been addressed.  Nice work!

README.md
```diff
@@ -18,6 +23,115 @@ Utility functions for Tetra Task Scripts
...
+|                            | h        | 1, 2, 3 ... 11, 12                      |
+| **Minute**                 | mm       | 00, 01, 02 ... 58, 59                   |
+|                            | m        | 0, 1, 2 ... 58, 59                      |
+| **Second**                 | ss       | 00, 01, 02 ... 58, 59                   |
+|                            | s        | 0, 1, 2 ... 58, 59                      |
+| **Fractional Second**      | S        | 0 1 ... 8 9                             |
+|                            | SS       | 00, 01, 02 ... 98, 99                   |
+|                            | SSS      | 000 001 ... 998 999                     |
+|                            | SSSS ... | 000[0..] 001[0..] ... 998[0..] 999[0..] |
+|                            | SSSSSS   |                                         |
```
mlipschultz, line 124: I'm confused by these two lines.
- The first shows four `S`es, but the output only shows three required digits (unless I'm misinterpreting the brackets).
- What does the ellipsis (`...`) mean after the `S`es?  That I can put more than four `S`es?
- The `SSSSSS` token doesn't have any example output.

__tests__/unit/test_convert_to_iso.py
```diff
@@ -0,0 +1,27 @@
...
+    PipelineConfig,
+)
+
+config_with_fold_test_cases = {
+    ("2021-11-07T01:30:00 America/New_York", 0, "2021-11-07T05:30:00.000Z"),
+    ("2021-11-07T01:30:00.032001 America/New_York", 1, "2021-11-07T06:30:00.032Z"),
+    ("2021-11-07T01:30:00.032 America/New_York", 1, "2021-11-07T06:30:00.032Z"),
+    ("2021-11-07T01:30:00.032", None, "2021-11-07T01:30:00.032"),
+
+    #TODO: Confirm these cases
```
mlipschultz, line 13: Did you ever confirm these cases?

task_script_utils/datetime_parser/datetime_info.py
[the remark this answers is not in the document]
mlipschultz: I think it's safe to assume military time if AM/PM isn't detected.  I'm not sure if parsing 13:00 AM as 13:00 PM or 1:00 PM is safe though.  I'll defer to @robjhornby-ts's opinion here since he's more familiar with the problem.

task_script_utils/datetime_parser/parser_exceptions.py
```diff
@@ -0,0 +1,28 @@
+class DatetimeParserError(Exception):
+    pass
+class InvalidYearError(DatetimeParserError):
```
mlipschultz, line 3: You're missing an empty line here.  I think there are supposed to be [two blank lines between classes](https://www.python.org/dev/peps/pep-0008/#blank-lines) though.

mlipschultz (changes requested):
I really like the README!

task_script_utils/datetime_parser/parser_exceptions.py
setiat: fixed

__tests__/unit/test_convert_to_iso.py
setiat: Nope, I couldn't find much examples of how `fold` works.
Need to build more test cases. It works for examples in confluence doc

README.md
setiat: I couldn't find any documentation for `...` in pendulum docs/ test cases.
May be it is  their shorthand to say that it works for 4,5 and 6 fraction digits as well.
I have updated the doc. It works for at max 6 digits.

```python
>>> import pendulum as p
>>> a =p.now()
>>> a.format("SSSSSS")
'105620'
>>> a.format("SSSSS")
'10562'
>>> a.format("SSSS")
'1056'
>>> a.format("SSS")
'105'
>>> a.format("SS")
'10'
>>> a.format("S")
'1'
```

README.md
```diff
@@ -18,6 +23,115 @@ Utility functions for Tetra Task Scripts
...
+    "PDT": "-07:00",
+    "PST": "-08:00",
+  }
+```
+
+- `fold`: `0` or `1`
+
+### Datetime format resolution matrix
+
+| DayFirst(down)/ Year_first | True     | False                  | None                                 |
```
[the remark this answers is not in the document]
setiat: fixed

__tests__/unit/test_datetime_parser.py
```diff
@@ -0,0 +1,140 @@
...
+}
+
+prase_with_no_format_list_test_cases = {
+    **format_list_test_cases,
+    "1:2:32 2021-12-23 AM America/Chicago": "2021-12-23T01:02:32-06:00",
+    "1:2:32 2021-12-23 AM +530": "2021-12-23T01:02:32+05:30",
+    "May 26 2013 12:12:12 AM Asia/Kolkata": "2013-05-26T00:12:12+05:30",
+    "Sunday, 26 2013 12:12:12 AM Asia/Kolkata": None,
+
+    #Error Case
```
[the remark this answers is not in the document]
setiat: added

task_script_utils/datetime_parser/datetime_info.py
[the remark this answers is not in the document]
setiat: The logic was updated as per test cases in google docs

task_script_utils/datetime_parser/datetime_info.py
```diff
@@ -0,0 +1,562 @@
...
+            raise AmbiguousDateError(
+                f"Can't decide day and month between: {tokens}")
+
+        day, month = (
+            (tokens[0], tokens[1])
+            if second_token_is_month
+            else (tokens[1], tokens[0])
+        )
+        day = f"0{day}" if day < 10 else str(day)
+        month = f"0{month}" if month < 10 else str(month)
```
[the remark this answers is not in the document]
setiat: fixed

task_script_utils/datetime_parser/datetime_info.py
```diff
@@ -0,0 +1,562 @@
...
+                    return f"{sign}{offset}"
+        return None
+
+    def match_am_or_pm(self, token: str):
+        """
+        Use regex to check if input string contains
+        AM or PM. Return the matched value
+        """
+        pattern = r"\s*[aApP][mM]\s+|\s*[aApP][mM]$"
+        matches = re.findall(pattern, token)
```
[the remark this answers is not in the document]
setiat: fixed.

task_script_utils/datetime_parser/date_formats.py
```diff
@@ -0,0 +1,52 @@
+from datetime import time
+from itertools import product
+
+from pydash.arrays import flatten
+
+date_parts = [
+    ['dddd', 'ddd ', 'dddd, ', 'ddd, ', ''],
+    ['MMMM', 'MMM'],
+    ['Do,', 'Do', 'D,', 'D'],
```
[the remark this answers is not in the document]
setiat: The logic has changed.  The the date format is built by looking at the datetime string and trying to find out which formatting token should be used. `DatetimeInfo.long_date_format` return the date format. Then a Cartesian product with all possible time formats is done to build a list of possible datetime formats.

This reduces the formatting list to 192 from 6400

task_script_utils/datetime_parser/datetime_info.py
```diff
@@ -0,0 +1,562 @@
...
+            are matched
+            - InvalidTimeError: When time like string is parsed, but the
+            numeric value of hrs, minutes and seconds are out of bound.
+
+        Returns:
+            Union[str, None]: return if any time string is found
+            else return None
+        """
+        hh_mm_ss_pattern = r"\d+:\d+:\d+.\d+|\d+:\d+:\d+"
+        hh_mm_pattern = r"\d{1,2}:\d{1,2}[+-]|\d{1,2}:\d{1,2}\s+|\d{1,2}:\d{1,2}$"
```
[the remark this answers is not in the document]
setiat: When the `seconds` are missing from time, `time` and offset will look the same. [+-] is used to distinguish between time and offset in this case.

task_script_utils/datetime_parser/datetime_info.py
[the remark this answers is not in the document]
robjhornby-ts: I've never seen a case like this so I think we can form an opinion. It should be rare.

My opinion is
- `13:00 AM` should be ambiguous, something like "AM/PM doesn't match time of day"
- `13:00 PM` - I wouldn't be surprised if this is produced by some software, but I haven't seen it so I'm not sure. Really, the "PM" is correct, it's just using a 24 hour clock and giving superfluous extra information that 13:00 is after midday. I think we can accept it but it's a grey area.
- If there is no AM/PM, assume it's a 24 hour format

task_script_utils/datetime_parser/datetime_info.py
```diff
@@ -0,0 +1,788 @@
...
+            token (str): A string value from `self.date_time_raw`
+            when splitted by whitespace
+
+        Returns:
+            str : An IANA timezone string
+        """
+        if token in pendulum.timezones:
+            return token
+
+    def _match_time(self, token: str) -> Union[str, None]:
```
mlipschultz, line 115: You'll need to update this and other methods' type hints for the return type.

mlipschultz (changes requested):

task_script_utils/datetime_parser/convert_to_iso_format.py
```diff
@@ -0,0 +1,42 @@
...
+
+import pendulum
+
+from .datetime_config import DatetimeConfig, DEFAULT_DATETIME_CONFIG
+from .parser import parse
+
+
+def convert_to_ts_iso8601(
+    datetime_str: str,
+    formats_list: Tuple[str] = (),
```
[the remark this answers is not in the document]
setiat: fixed.

task_script_utils/datetime_parser/datetime_info.py
setiat: fixed.

task_script_utils/datetime_parser/datetime_info.py
```diff
@@ -0,0 +1,562 @@
...
+            - MultipleTimesFoundError: When more than one time-string
+            are matched
+            - InvalidTimeError: When time like string is parsed, but the
+            numeric value of hrs, minutes and seconds are out of bound.
+
+        Returns:
+            Union[str, None]: return if any time string is found
+            else return None
+        """
+        hh_mm_ss_pattern = r"\d+:\d+:\d+.\d+|\d+:\d+:\d+"
```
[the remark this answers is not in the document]
mlipschultz: Since it's been awhile, @setiat what are your plans for this comment?

task_script_utils/datetime_parser/datetime_info.py
```diff
@@ -0,0 +1,788 @@
...
+        for key_, values in token_map.items():
+            if token in values:
+                return key_
+            elif token.replace(",", "") in values:
+                return f"{key_},"
+        return None
+
+    @property
+    def date_str(self):
+        """Returns YYYY-MM-DD"""
```
mlipschultz, line 374: Not necessarily.  There are code paths that can result in single-digit months and days, meaning this method may also return `YYYY-M-D`.  Can you update the code so that this method (and others, where appropriate) always return date as `YYYY-MM-DD`?  Can you also add tests to confirm this behavior?

To confirm this, I added `("2021.11.7", None, False,'2021-11-07'),` to `two_digit_date_with_config_test_cases` in the `datetime_info.py` test file.  I was suspicious that `_process_year_first_or_last_matches` wasn't doing any padding to the days and months.

task_script_utils/datetime_parser/convert_to_iso_format.py
```diff
@@ -0,0 +1,42 @@
...
+
+from .datetime_config import DatetimeConfig, DEFAULT_DATETIME_CONFIG
+from .parser import parse
+
+
+def convert_to_ts_iso8601(
+    datetime_str: str,
+    formats_list: Sequence[str] = (),
+    config: DatetimeConfig = DEFAULT_DATETIME_CONFIG
+):
```
mlipschultz, line 13:
```suggestion
) -> str:
```

__tests__/unit/test_convert_to_iso.py
mlipschultz: What do you need to confirm these?  @robjhornby-ts, would you be able to help out here?

mlipschultz (changes requested):

__tests__/unit/test_convert_to_iso.py
robjhornby-ts: Sorry for taking a while to get to this. These ones are correct. I'd add these which already work as expected with the current code:

```
    # On a date when the fold is irrelevant, it is ignored
    ("2021-11-06T01:30:00 America/New_York", 0, "2021-11-06T05:30:00.032Z"),
    ("2021-11-06T01:30:00 America/New_York", 1, "2021-11-06T05:30:00.032Z"),
    ("2021-11-08T01:30:00 America/New_York", 1, "2021-11-08T06:30:00.032Z"),
    ("2021-11-08T01:30:00 America/New_York", 0, "2021-11-08T06:30:00.032Z"),

    # Other timezones
    ("2021-10-31T01:15:00 Europe/London", 1, "2021-10-31T01:15:00Z"),
    ("2021-10-31T01:15:00 Europe/London", 0, "2021-10-31T00:15:00Z"),
    ("2021-10-31T02:45:00 Europe/Rome", 1, "2021-10-31T01:45:00Z"),
    ("2021-10-31T02:45:00 Europe/Rome", 0, "2021-10-31T00:45:00Z"),
```

We'll need to change the logic slightly for when `fold=None` - it should raise an error like AmbiguousFoldError if `fold=None` during a time when `fold` is required to disambiguate the datetime. These cases should raise an error. Each one is `(input_, fold)`

```
    # Raises AmbiguousFoldError
    ("2021-11-07T01:30:00 America/New_York", None),
    ("2021-11-07T01:20:00.03 America/New_York", None),
    ("2021-10-31T01:15:00.123 Europe/London", None),
    ("2021-10-31T02:45:00.0 Europe/Rome", None),
```

The reason these are ambiguous is because their UTC time could be one of two different values depending on whether fold = 0 or 1. A simple way to check using the existing code could be something like this, assuming a TSDatetime has already been created:

```python
import copy


def is_fold_required(datetime_: TSDatetime) -> bool:
    """Check whether an undefined `fold` would cause an ambiguous TSDatetime

    The "fold" is required during the 2 hour window when clocks are set back in
    a timezone which keeps track of daylight savings (such as IANA timezones
    like `Europe/London`).

    This function returns True if "fold" is required to disambiguate a
    datetime, False otherwise.
    """
    # Copy because TSDatetime is mutable
    dt = copy.deepcopy(datetime_)
    dt.change_fold(0)
    dt_before_fold = dt.ts_format
    dt.change_fold(1)
    dt_after_fold = dt.ts_format

    return dt_before_fold != dt_after_fold

if fold is None and is_fold_required(datetime_):
    raise AmbiguousFoldError(
        "This time is during the transition from daylight savings to standard "
        f"time: {datetime_input_string}. To disambiguate it, we need to know "
        "whether it is 'before the fold' (during daylight savings time) or "
        "'after the fold' (during standard time)."
    )
```

There might be a simpler way too (e.g. I'm not sure if I need to use `dt.ts_format` for the comparison, might be overkill)

README.md
```diff
@@ -18,6 +23,116 @@ Utility functions for Tetra Task Scripts
...
+    "CDT": "-05:00",
+    "CST": "-06:00",
+    "MDT": "-06:00",
+    "MST": "-07:00",
+    "PDT": "-07:00",
+    "PST": "-08:00",
+  }
+```
+
+- `fold`: `0` or `1`
```
robjhornby-ts, line 73: Might want to add an explanation too, hopefully this isn't too long but people seem to find "fold" confusing without giving all the detail.

Fold is relevant for timezones which refer to both a daylight savings time and a standard time, such as IANA timezones like `America/New_York` which is `-05:00` in standard time and `-04:00` in daylight savings time. For these timezones, there is a period around the transition from daylight savings to standard time where the same time appears on the clock twice, an hour apart. Without knowing whether it's the first time (fold=0) or the second time (fold=1), the datetime is ambiguous when converting to UTC.

For example, on the date when daylight savings end in New York, "2021-11-07T**01**:30:00 **America/New_York**" could refer to either "2021-11-07T**05**:30:00**Z**" or "2021-11-07T**06**:30:00**Z**" depending on the fold. Supplying a `fold` of `0` or `1` disambiguates this. However, `fold` is typically not provided with raw data using this format: if `fold` is required to convert to UTC reliably, but isn't included in the datetime config, an error will be raised. See [PEP 495](https://www.python.org/dev/peps/pep-0495/) more on disambiguating local time.

robjhornby-ts (commented):
A couple of comments

I think we found that some places in the world use a YYYY/DD/MM date format so we couldn't assume that `AAAA BB CC` is always Y M D. But it seems odd that `parse(datetime_str="2021-01-02T01:30:00Z")` requires a `day_first` `False` config, even though it follows the RFC3999 format and our own TS ISO format. It feels like the default behaviour should be that it accepts a format of `YYYY-MM-DDTHH:mm:ssZ` (with or without microseconds, optionally with a space instead of the `T`, and with any kind of supported timezone, not just the `Z` token) as unambiguous. 

Could we potentially do this by defining our default `formats_list` to contain some formats like above, instead of an empty tuple? Or maybe create a tuple of formats which can be passed to the formats_list input like `parse(datetime_str="2021-01-02T01:30:00Z", formats_list=DEFAULT_FORMATS)` (maybe with a better name)

If any of these comments would take a long time to implement, I'm also happy to leave them as future improvements instead of prolonging this PR

task_script_utils/datetime_parser/datetime_info.py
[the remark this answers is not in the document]
setiat: updated the regex pattern, they are not exactly the same as suggested. But they cover the test cases

task_script_utils/datetime_parser/datetime_info.py
```diff
@@ -0,0 +1,788 @@
...
+            when splitted by whitespace
+
+        Raises:
+            - MultipleTimesFoundError: When more than one time-string
+            are matched
+            - InvalidTimeError: When time like string is parsed, but the
+            numeric value of hrs, minutes and seconds are out of bound.
+
+        Returns:
+            Union[str, None]: return if any time string is found
```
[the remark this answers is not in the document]
setiat: fixed

task_script_utils/datetime_parser/datetime_info.py
setiat: Updated the doc string... The idea is to return raw values as year-month-day

__tests__/unit/test_datetime_info.py
```diff
@@ -0,0 +1,132 @@
...
+
+
+@pytest.mark.parametrize("input, year_first, day_first, expected", two_digit_date_with_config_test_cases)
+def test_match_short_date(input, year_first, day_first, expected):
+    config_dict = datetime_configs[(year_first, day_first)]
+    config = DatetimeConfig(**config_dict)
+    try:
+        date_info = DateTimeInfo(input, config)
+        result = date_info.date_str
+    except Exception as e:
```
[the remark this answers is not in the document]
setiat: fixed.

task_script_utils/datetime_parser/datetime_info.py
```diff
@@ -0,0 +1,788 @@
...
+            token (str): A string value from `self.date_time_raw`
+            when splitted by whitespace
+
+        Returns:
+            Union[str, None]: if a match occurs return DD-MM-YYYY formated
+            date string else return None
+        """
+        year_first_pattern = r"\d{4,4}[-./\\]\d{1,2}[-./\\]\d{1,2}"
+        year_last_pattern = r"\d{1,2}[-./\\]\d{1,2}[-./\\]\d{4,4}"
+        two_digit_date_pattern = r"^\d{1,2}[-./\\]\d{1,2}[-./\\]\d{1,2}"
```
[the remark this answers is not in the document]
setiat: fixed

__tests__/unit/test_datetime_info.py
```diff
@@ -0,0 +1,135 @@
...
+]
+
+
+@pytest.mark.parametrize("input, year_first, day_first, expected", two_digit_date_with_config_test_cases)
+def test_match_short_date(input, year_first, day_first, expected):
+    config_dict = datetime_configs[(year_first, day_first)]
+    config = DatetimeConfig(**config_dict)
+    try:
+        date_info = DateTimeInfo(input, config)
+        result = date_info.date_str
```
mlipschultz, line 114: @robjhornby-ts: Continuing the conversation [from another thread](https://github.com/tetrascience/ts-task-script-utils/pull/2#discussion_r762754313), if the input string is `"2021.11.7"`, is the correct output:
1. `'2021-11-07'`
2. `'2021-11-7'`
3. Both would be considered valid

robjhornby-ts: The only place I see this property used is in this test, not in the logic for creating the TS ISO 8601 string. Can we delete the `date_str` and `time_str` properties of the DateTimeInfo class and change these tests to use other properties of the DateTimeInfo object instead?

@tariksetia the DateTimeInfo class has a lot of methods and is relatively complex already, we could remove everything which isn't used elsewhere in the parsing logic.

mlipschultz: Good catch Rob!  I didn't realize `date_str` was only used in the tests.  I agree with removing the methods only used in testing.  I'd still like to keep the test cases around though, so Tarik can you update the tests for `convert_to_ts_iso8601` to also include these test cases?

setiat: `date_str` and `time_str` are required for internal processing when parsing date and time.

[line 71](https://github.com/tetrascience/ts-task-script-utils/blob/DE-2591-datetime-parser/task_script_utils/datetime_parser/datetime_info.py#L71) and [line 96](https://github.com/tetrascience/ts-task-script-utils/blob/DE-2591-datetime-parser/task_script_utils/datetime_parser/datetime_info.py#L97)

setiat: - `DateTimeInfo` is an internal data structure. Not to be used by developer. So wether it is '2021-11-07' or `'2021-11-7'`, should not matter at DateTimeInfo level
- Developer can use only `convert_to_ts_iso8601` and `parse`
-`convert_to_ts_iso8601` uses `parse()` internally
- Now `parse` returns `TSDatetime` which exposes
      - `TSDatetime.datetime`: return python datetime object
      -  `TSDatetime.iso_format`: return date time str with 2 digit months and days
      -  `TSDatetime.ts_format`: return date time str with 2 digit months and days

mlipschultz: > `date_str` and `time_str` are required for internal processing when parsing date and time.

The way the code's written, it's not possible for us to make use of the code tracing tools that IDEs like PyCharm and VS Code provide, making maintenance of the code difficult.  For example, finding usages of the `date_str` property doesn't find line 97 because that's just a string and the IDE doesn't know that `"date_str"` refers to the property `date_str`.  I'd like to see the code rewritten to enable this kind of tracing, but I think that's another issue.

> Developer can use only `convert_to_ts_iso8601` and `parse`

This is why I'm thinking we need to have more tests on these functions.  For example, there are no tests confirming that they always return 2-digit months and days.  The only tests at this level are for fold and fractional seconds.  It would be good to have some smoke tests just to confirm that everything works as expected.  I think the smoke tests should include:
* the cases in [the datetime_dayfirst_yearfirst spreadsheet](https://docs.google.com/spreadsheets/d/1GM_rbbYWab9h_NIs-WKLAdHvh-zJQqDpRgb0buk2EG0/edit#gid=0)
* are month and days always digits long?

I know you have tests for this already, but they're on methods that are known to not produce the correct output (e.g. `date_str` on single-digit months and days).  That's why I'm asking for smoke tests on the front-end functions developers are expected to use.  Having these tests on the front-end will also create a safety net if/when we update the backend to my concern above about maintainability.

README.md
setiat: added

__tests__/unit/test_datetime_info.py
setiat: added tests for padding  and configured padding for date and time parts in logic.
If it is not urgent can I take parser modification as  a separate task for later sprints along?

task_script_utils/datetime_parser/datetime_info.py
setiat: added padding to  all datetime parts

mlipschultz (commented):
I talked with Kai and we realized it'd be better to have a single function as the entrypoint to the parsing.  We think it should be the `parse` function since you can use `parse(...).tsformat()` to get a TS-formatted string.  To accomplish this, can you:
1. Add a docstring to the `parse` function (the one for `convert_to_ts_iso8601` would work with some minor changes)
3. Rename `iso_format` to isoformat` and make it a method.  This is so it's consistent with [Python's datetime object method](https://docs.python.org/3/library/datetime.html#datetime.datetime.isoformat) -- reduces cognitive load on developers.
2. Rename `ts_format` to `tsformat` and make it a method.  This is to be consistent with `isoformat` above.

Eventually, we will remove `convert_to_ts_iso8601` but in the interest of time we'll leave it for now.  This also means that for now we can keep the tests on `convert_to_ts_iso8601` instead of moving them to `parse`.  I've also added this to the future issue.

mlipschultz:
@robjhornby-ts, @wkopen: Should we be able to parse just dates or just times, e.g.:
```python
parse('2021-12-08')
parse('12:49:51')
```
Or should those produce errors?

wkopen:
> @robjhornby-ts, @wkopen: Should we be able to parse just dates or just times, e.g.:
> 
> ```python
> parse('2021-12-08')
> parse('12:49:51')
> ```
> 
> Or should those produce errors?

Yes I think there are cases where only date or time is provided. It's not a common case though. Let's make a separate ticket for it and document in the README it's a limitation. Not required in this version.

__tests__/unit/test_datetime_parser.py
```diff
@@ -0,0 +1,159 @@
...
+    "dddd, MMMM Do, YYYY hh:mm:ss.SSSSSS A z",
+
+]
+
+tz_dict = {
+    "IST": "+05:30",
+    "BST": "+01:00"
+}
+
+format_list_test_cases = {
```
wkopen, line 27: is `formatted_list_test_cases` better? just `format` makes it sound like a verb. If you agree, then please update the others

task_script_utils/datetime_parser/datetime_config.py
```diff
@@ -0,0 +1,31 @@
...
+        tz_dict={},
+        fold=None
+    ):
+
+        self.day_first = day_first
+        self.year_first = year_first
+        self.tz_dict = tz_dict
+        self.tz_dict_seconds = map_offset_to_seconds(tz_dict)
+
+        # The "fold" is required during the 2 hour window when clocks are set back in
```
wkopen, line 18: This is good but I think docstring is a more standard way to document this

__tests__/unit/test_datetime_parser.py
```diff
@@ -0,0 +1,159 @@
...
+    "Sunday, May 26th 2013 12:12:12 AM BST": "2013-05-26T00:12:12+01:00",
+    "Sunday, May 26th 2013 12:12:12 AM Asia/Kolkata": "2013-05-26T00:12:12+05:30",
+    "Sunday, May 26th, 2013 12:12:12 AM Asia/Kolkata": "2013-05-26T00:12:12+05:30",
+    "Sunday, May 26 2013 12:12:12 AM Asia/Kolkata": "2013-05-26T00:12:12+05:30",
+    "Sunday, May 26 2013 12:12:12.5677 AM Asia/Kolkata": "2013-05-26T00:12:12.5677+05:30",
+    "Sunday, May 26th 2013 12:12:12.5677 AM Asia/Kolkata": "2013-05-26T00:12:12.5677+05:30",
+}
+
+parse_with_no_format_list_test_cases = {
+    **format_list_test_cases,
```
wkopen, line 39: We are in `parse_with_no_format_list_test_cases` but you are adding formatted test cases. It's inconsistent with the name of this test case.

__tests__/unit/test_datetime_info.py
```diff
@@ -0,0 +1,136 @@
...
+    ("2021/11/07", None, None, None),
+    ("2021/32/07", None, True, None),
+    ("2021/11/14", None, True, None),
+    ("2021.11.7", None, False,'2021-11-07')
+]
+
+# Test them with year_first
+# The goal is to test regex
+regex_test_cases = [
+    ("1:2:32 20-12-1 AM America/Chicago", "01-12-2020 01:02:32 AM America/Chicago"),
```
wkopen, line 77: I'm not sure I understand this. I assume the original string means YY-MM-DD. But in your expected string it's DD-MM-YYYY? What's the purpose of converting it from a non-confusing form to a confusing form?

__tests__/unit/test_datetime_parser.py
```diff
@@ -0,0 +1,159 @@
+import pytest
```
wkopen, line 1: I assume `test_datetime_parser.py` is more like a unit test but this file is more like an integration test. I'm looking for more formats covered in integration tests actually just like what you had in unit tests.

And, if this file serves more like a integration test, we should probably move it to `__tests__/integration`. If this file is not an integration test, I'm looking for an integration (end-to-end) test file. In the integration test, if you can also include the format with `Z` in the end that will be great because in most cases people will be using the UTC+0 ISO format in the IDS JSON.

wkopen (changes requested):
I only reviewed the test cases. Will keep reviewing tomorrow.

__tests__/unit/test_datetime_parser.py
setiat: renamed it to `datetime_formats_list_test_cases`

task_script_utils/datetime_parser/datetime_config.py
setiat: added doc strings

__tests__/unit/test_datetime_parser.py
setiat: Fixed it 
and renamed the variable to `parse_with_no_datetime_formats_list_test_cases`

__tests__/unit/test_datetime_info.py
setiat: Fixed DatetimeInfo code to always return date in Year-month-day

setiat:
> > @robjhornby-ts, @wkopen: Should we be able to parse just dates or just times, e.g.:
> > ```python
> > parse('2021-12-08')
> > parse('12:49:51')
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Or should those produce errors?
> 
> Yes I think there are cases where only date or time is provided. It's not a common case though. Let's make a separate ticket for it and document in the README it's a limitation. Not required in this version.

added as a limitation to readme

task_script_utils/datetime_parser/ts_datetime.py
```diff
@@ -26,23 +26,23 @@ def tzinfo(self):
...
     def datetime(self):
-        if self._subseconds is None:
-            return self._datetime
-
-        microseconds = int(str(self._subseconds)[:6])
+        """Return a new `datetime` object after replacing `microseconds`
+        with `self._subseconds` in `self._datetime`.
+        `subseconds` needs to be truncated to 6 digits
+        """
+        microseconds = self._subseconds[:6]
```
mlipschultz, line 33: Won't this fail with a TypeError if `self._subseconds` is None?  From the `__init__`, it looks like that's an expected value for `self._subseconds`.

mlipschultz (changes requested):
Just one fairly small request, then I think it should be good to merge.

__tests__/integration/test_convert_to_iso.py
```diff
@@ -0,0 +1,159 @@
...
+    ("01/02/3T04:30:00 America/New_York", False, True, "2003-02-01T09:30:00Z"),
+    ("1/2/3T4:30:00 America/New_York", False, True, "2003-02-01T09:30:00Z"),
+    ("1/2/3T4:3:00 America/New_York", False, True, "2003-02-01T09:03:00Z"),
+    ("01/02/13T04:03:00 America/New_York", False, True, "2013-02-01T09:03:00Z"),
+    ("01/02/03 04:03:00 America/New_York", True, True, "2001-03-02T09:03:00Z"),
+    ("01/02/03 04:03:00 America/New_York", False, True, "2003-02-01T09:03:00Z"),
+    ("01/02/03 04:03:00 America/New_York", None, True, "2003-02-01T09:03:00Z"),
+    ("1/02/03 04:03:00 America/New_York", True, False, "2001-02-03T09:03:00Z"),
+    ("01/02/03 04:03:00 America/New_York", False, False, "2003-01-02T09:03:00Z"),
+    ("01/02/03 04:03:00 America/New_York", None, False, None),
```
wkopen, line 62: Please see my suggestion in README on how to organize test cases

__tests__/integration/test_convert_to_iso.py
```diff
@@ -0,0 +1,159 @@
...
+    ("01/15/11 04:03:00 America/New_York", None, True, None),
+    ("01/15/11 04:03:00 America/New_York", True, False, None),
+    ("12/13/03 04:03:00 America/New_York", None, False, "2003-12-13T09:03:00Z"),
+    ("2021.11.07 04:03:00.00045000 America/New_York",
+     None, False, "2021-11-07T09:03:00.00045000Z"),
+
+    ("2021/11/07 04:03:00 America/New_York", None, True, "2021-07-11T08:03:00Z"),
+
+    ("11\\12\\2021 04:03:00 America/New_York", None, True, "2021-12-11T09:03:00Z"),
+    ("2021/11/07 04:03:00 America/New_York", None, None, None),
```
wkopen, line 85: I'm curious why we can't parse this. I've never seen in any date format YYYY/DD/MM. I don't think it exists. Theoretically it can, but in reality it never does. Please see this wikipedia: https://en.wikipedia.org/wiki/Date_format_by_country#

README.md
```diff
@@ -18,6 +24,125 @@ Utility functions for Tetra Task Scripts
...
+
+### DatetimeConfig
+
+```python
+from task_script_utils.datetime_parser import DatetimeConfig
+```
+
+A `DatetimeConfig` object has following attributes:
+
+- `day_first`: `True/False/None`
```
wkopen, line 48: Please specify if it's optional or required and if optional what's the default value. It's very standard in API docs. I think default is `None` but better to specify.

README.md
```diff
@@ -18,6 +24,125 @@ Utility functions for Tetra Task Scripts
...
+- `year_first`: `True/False/None`
```
wkopen, line 49: We should explicitly say if `year_first` is `True`, then we assume the format will be YMD because there is no YDM format in any country. It's a very safe assumption and backed by https://en.wikipedia.org/wiki/Date_format_by_country#

README.md
```diff
@@ -18,6 +24,125 @@ Utility functions for Tetra Task Scripts
...
+
+- `fold`: `0`, `1` or `None`.
+  - It is required during the 2 hour window when clocks are set back in a timezone which keeps track of daylight savings (such as IANA timezones like `Europe/London`).
+  - The allowed values for the fold attribute will be 0 and 1 with 0 corresponding to the earlier and 1 to the later of the two possible readings of an ambiguous local time.
+  - If fold is `None`, Parser will check if `fold` is needed or not to parse the time with no ambiguity.
+  - `AmbiguousFoldError` will be raised if `fold` is needed.
+
+### Datetime format resolution matrix
+
+| day_first(down)/ year_first | True     | False                  | None                                 |
```
wkopen, line 82: This matrix is confusing because of the way it's formatted. I think it's better to just have `day_first` `year_first` on different columns and `resolution` column. It's also better to use single character like YMD instead of YY-MM-DD. I assume you just want to present the order instead of format. YY-MM-DD in some context is a format and it also implies two-digit year value.

README.md
```diff
@@ -18,6 +24,125 @@ Utility functions for Tetra Task Scripts
...
+| --------------------------- | -------- | ---------------------- | ------------------------------------ |
+| **True**                    | YY-DD-MM | DD-MM-YY               | DD-MM-YY                             |
```
wkopen, line 84: YDM order never exists. This is a safe assumption and should be documented.

README.md
```diff
@@ -18,6 +24,125 @@ Utility functions for Tetra Task Scripts
...
+```python
+from task_script_utils.datetime_parser import convert_to_ts_iso8601
+
+convert_to_ts_iso8601("2004-12-23T12:30 AM +05:30")
+convert_to_ts_iso8601("2004-12-23T12:30 AM +05:30", <datetime_config>)
+convert_to_ts_iso8601("2004-12-23T12:30 AM +05:30", <format_list>)
+convert_to_ts_iso8601("2004-12-23T12:30 AM +05:30", <format_list>, <datetime_config>)
+```
+
+### DatetimeConfig
```
wkopen, line 40: * I would say most people don't care about this config. They just want our date time conversion to work. So before we introduce this config, we should list out all unambiguous formats which doesn't need any config. And we should also list out all ambiguous formats that will require this additional config otherwise we can't parse it.
* This should also determine how we organize our test cases. The two main parts should be "unambiguous" and "ambiguous". Then in "ambiguous", we should group by date time format. For example, you want to test `21-12-12T13:00:00`, it's nice to do none of the config defined, then move on to only `year_first` is defined, only `day_first` is defined, only `tz_dict` is defined. Then move to 2 of the configs defined, then move to all 3 defined. This way, the tests are organized in a permutation order.
* If the date time is already "unambiguous", will config do anything to it? It's also worth documenting the behavior.

__tests__/unit/test_datetime_info.py
```diff
@@ -0,0 +1,106 @@
...
+from task_script_utils.datetime_parser.datetime_info import DateTimeInfo
+from task_script_utils.datetime_parser.datetime_config import DatetimeConfig
+from task_script_utils.datetime_parser import tz_dicts
+from task_script_utils.datetime_parser.parser_exceptions import DatetimeParserError
+
+
+two_digit_date_with_config_test_cases = [
+    # input, year_first, day_first, expected
+    # Expected date format = YYYY-MM-DD
+    ("01/02/03", True, True, "2001-03-02"),
```
wkopen, line 12: Please see my comments in the README about how to order the test cases. It's very hard to follow right now.

__tests__/integration/test_convert_to_iso.py
```diff
@@ -0,0 +1,159 @@
...
+    ("2021/32/07 04:03:00 America/New_York", None, True, None),
+    ("2021/11/14 04:03:00 America/New_York", None, True, None),
```
wkopen, line 87: This is also my question in README, what's our behavior if the date time is already unambiguous. It's important to document the behavior.

__tests__/integration/test_convert_to_iso.py
```diff
@@ -0,0 +1,159 @@
...
+    ("13/02/03 04:03:00 America/New_York", True, True, "2013-03-02T09:03:00Z"),
+    ("13/2/03 04:03:00.43500 America/New_York",
+     False, True, "2003-02-13T09:03:00.43500Z"),
+    ("13/02/03 04:03:00 America/New_York", None, True, "2003-02-13T09:03:00Z"),
+    ("13/02/03 04:03:00 +05:30", True, False, "2013-02-02T22:33:00Z"),
+    ("13/02/3 04:03:00 America/New_York", False, False, None),
+    ("13-02-03 04:03:00 America/New_York", None, False, "2013-02-03T09:03:00Z"),
+    ("13/02/03 04:03:00 America/New_York", True, None, "2013-02-03T09:03:00Z"),
+    ("13/2/03 04:03:00 America/New_York", False, None, "2003-02-13T09:03:00Z"),
+    ("13/02/03 04:03:00 America/New_York", None, None, None),
```
wkopen, line 75: This is an unambiguous format. 13 can never be a month value, so it's implicit that it's DDMMYY. I think countries use 2-digit year in the beginning is very rare according to Wiki (probably just Kenya). So we should make it a safe assumption.

__tests__/integration/test_convert_to_iso.py
```diff
@@ -0,0 +1,159 @@
...
+    ("01/15/11 04:03:00 America/New_York", None, True, None),
```
wkopen, line 76: This is also an unambiguous format, 15 can never be a month value, so it's implicit that the format is MMDDYY. So we should be able to parse it. Again, all of this is due to we don't have a clear spec on what's our parser expected behavior

wkopen (changes requested):
Hi @setiat , I think the most important observation from reviewing the test cases is "we don't have a clear spec doc on the parser behavior under different situations". You can see my specific comments in the README.md file. Without a clear spec doc, nobody can clearly tell what the parser will do under certain situations. It's not completely on you to be fair. The Confluence page we have is also not documented clearly.

So I recommend starting from the README.md. Just write out the spec doc as I suggested, then return to me for another review before you change any code. Once the spec doc is good, you can go ahead modifying the code and tests. Let's don't rush it. I know this task has been worked on for a while. We (Rob, Mike and I) take the responsibility of not giving you a much earlier review on the design.

README.md
robjhornby-ts: That just means people should almost never define `day-first` `True` & `year-first` `True` in their datetime config; but it is possible and this is the resulting format

__tests__/integration/test_convert_to_iso.py
robjhornby-ts: I've seen 2 digit years first in a plenty of software even if it's not an official country format, e.g. it's one of the Windows date formats, which means Empower data could have this format depending on the settings of the PC generating it. (That makes this an ambiguous format)

includes `yy/MM/dd`:
image

__tests__/integration/test_convert_to_iso.py
robjhornby-ts: as above, could be YYMMDD

__tests__/integration/test_convert_to_iso.py
robjhornby-ts: Agreed on this one, 4 digit year first -> assume YYYYMMDD

__tests__/integration/test_convert_to_iso.py
wkopen: You are right Rob. Please ignore what I said here. It is ambiguous.

__tests__/integration/test_convert_to_iso.py
wkopen: But here we know it can only be MDY because YDM never exists.

README.md
wkopen: I agree

task_script_utils/datetime_parser/datetime_info.py
```diff
@@ -0,0 +1,807 @@
...
+            tuple: If date is parsed successfully return
+            (True, datetime object) else return (False, None)
+        """
+        try:
+            parsed = pendulum.from_format(date_str, format)
+            return parsed
+        except Exception as e:
+            return None
+
+    def _try_formats(self, date_str: str, formats: Tuple[str]):
```
mlipschultz, line 762: [minor] Just so the type checker knows it's ok to have more than one string in the tuple, although I think you don't need to require a `Tuple`, but rather a more generic type (`Iterable`?)

```suggestion
    def _try_formats(self, date_str: str, formats: Tuple[str, ...]):
```

task_script_utils/datetime_parser/datetime_info.py
```diff
@@ -0,0 +1,807 @@
...
+        """Given a date string and a list for formats, make sure
+        sure only one of the format successfully parses the date.
```
mlipschultz, line 764: [minor]
```suggestion
        """Given a date string and a list for formats, make sure
        only one of the format successfully parses the date.
```

mlipschultz (commented):
Two minor suggestions for improvement.

task_script_utils/datetime_parser/datetime_info.py
```diff
@@ -534,11 +496,7 @@ def long_date_format(self) -> str:
                 f"YYYY"
             )
         else:
-            date_fmt = (
-                f"{self.token_month} "
-                f"{self.token_day} "
-                f"YYYY"
-            )
+            date_fmt = f"{self.token_month} " f"{self.token_day} " f"YYYY"
```
mlipschultz, line 499: I'm guessing black auto-formatted this weird, but could you fix it to be more readable:
```suggestion
            date_fmt = f"{self.token_month} {self.token_day} YYYY"
```

task_script_utils/datetime_parser/datetime_info.py
```diff
@@ -603,9 +560,18 @@ def _process_two_digit_date_pattern(self, date_parts):
...
+                # Input = YY-MM-DD
+                year, month, day = date_parts
             else:
                 # Input = YY-MM-DD
                 year, month, day = date_parts
+                if int(month) > 12:
+                    month, day = day, month
+                # At this point both month and day
+                # could have improper values, eg day=42 and month=16
+                # This is validate later below in this function.
```
mlipschultz, line 573:
```suggestion
                # This is validated later below in this function.
```

task_script_utils/datetime_parser/datetime_config.py
```diff
@@ -8,21 +8,33 @@ def __init__(
...
         year_first: Optional[bool] = None,
         tz_dict: dict = {},
-        fold: Optional[int] = None
+        fold: Optional[int] = None,
     ):
-        """DatetimeConfig is used to provide complementary information that helps
-        to parse datetime
+        """DatetimeConfig provides complementary information on how to mark
+        parsed digits as day, month or year.
+        It also provide an options to handle abbreviated time zones and
```
mlipschultz, line 15:
```suggestion
        It also provides an option to handle abbreviated time zones and
```

task_script_utils/datetime_parser/datetime_config.py
```diff
@@ -8,21 +8,33 @@ def __init__(
...
+        fold for parsing ambiguous timestamps during daylight saving transitions.
+        Ideally, DatetimeConfig should be constructed from pipeline configuration
+        passed to task scripts
```
mlipschultz, line 18:
```suggestion
        passed to task scripts.
```

task_script_utils/datetime_parser/datetime_config.py
```diff
@@ -8,21 +8,33 @@ def __init__(
...
-            like `Europe/London`). Defaults to None.
+            day_first (Optional[bool], optional): Whether to interpret the first
+            value in an ambiguous 3-integer date (e.g. 01/05/09) as the
+            day (True) or month (False). Defaults to None.
+
+            year_first (Optional[bool], optional): Whether to interpret the first
+            value in an ambiguous 3-integer date (e.g. 01/05/09) as the year.
+            Defaults to None. When the year has four digits, then whether
+            `year_first` is `true` or `false`,is decided by regex parsing done
+            by `DatetimeInfo` class.
```
mlipschultz, line 29: Both `day_first` and `year_first` claim to interpret the first value as day or year.  Can you add something to the docstring to clarify what happens?  If they're both true, then `year_first` takes priority and `day_first` starts with the second value, right?

mlipschultz (commented):
I found some typos to fix and a case where black made a bad formatting decision.  Plus, it looks like there are still some unaddressed review comments from earlier.

task_script_utils/datetime_parser/datetime_config.py
setiat: fixed.

task_script_utils/datetime_parser/datetime_config.py
setiat: fixed,

task_script_utils/datetime_parser/datetime_config.py
setiat: Updated doc strings and Yes, the resulting format will be Y-D-M

task_script_utils/datetime_parser/ts_datetime.py
setiat: it was already fixed : https://github.com/tetrascience/ts-task-script-utils/blob/DE-2591-datetime-parser/task_script_utils/datetime_parser/ts_datetime.py#L28

Not sure why it is  not tagged as  outdated

README.md
```diff
@@ -26,122 +26,22 @@ Utility functions for Tetra Task Scripts
...
 ```python
-from task_script_utils.datetime_parser import DatetimeConfig
+from task_script_utils.datetime_parser import parse
+
+parse("2004-12-23T12:30 AM +05:30")
+parse("2004-12-23T12:30 AM +05:30", <datetime_config>)
+parse("2004-12-23T12:30 AM +05:30", <format_list>)
+parse("2004-12-23T12:30 AM +05:30", <format_list>, <datetime_config>)
+
+# Parse() returns a TSDatetime Object.
```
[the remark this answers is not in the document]
setiat: fixed. Moved it out of code block

task_script_utils/datetime_parser/datetime_info.py
setiat: fixed

task_script_utils/datetime_parser/datetime_info.py
setiat: fixed

setiat:
@mlipschultz To pass the PR checks, I have to restrict python version to 3.7. This applies to both travis build and GitHub hooks. Prior to black it was running on 3.7 and during its introduction, `requires` was removed from Pipfile. I could not make it work without restring the python version.

All other comments were also addressed.

task_script_utils/datetime_parser/datetime_config.py
mlipschultz: It's minor, but you deleted the `s` instead of adding the `.`

task_script_utils/datetime_parser/datetime_config.py
mlipschultz: You still need to add an `s` to `provide`

setiat: fixed 😅

task_script_utils/datetime_parser/datetime_config.py
setiat: fixed

mlipschultz (changes requested):
I went through all unresolved comments and resolved the ones that looked addressed (sorry Kai and Rob if you wanted to resolve your own comments).  I left a few more comments to address.

@wkopen, I'd like you to take a look at these comments in particular.  You had some questions about test cases vs. spec and I tried to follow the spec to confirm test case behavior.
* https://github.com/tetrascience/ts-task-script-utils/pull/2/files#r770240480
* https://github.com/tetrascience/ts-task-script-utils/pull/2/files#r770239247 -- I think this reveals a mismatch between the spec and the code

Since I couldn't respond to some comments, here are the comments any my questions:

https://github.com/tetrascience/ts-task-script-utils/pull/2#discussion_r770231421
> * I would say most people don't care about this config. They just want our date time conversion to work. So before we introduce this config, we should list out all unambiguous formats which doesn't need any config. And we should also list out all ambiguous formats that will require this additional config otherwise we can't parse it.
> * This should also determine how we organize our test cases. The two main parts should be "unambiguous" and "ambiguous". Then in "ambiguous", we should group by date time format. For example, you want to test `21-12-12T13:00:00`, it's nice to do none of the config defined, then move on to only `year_first` is defined, only `day_first` is defined, only `tz_dict` is defined. Then move to 2 of the configs defined, then move to all 3 defined. This way, the tests are organized in a permutation order.
> * If the date time is already "unambiguous", will config do anything to it? It's also worth documenting the behavior.

I don't know that we ever list out all unambiguous formats.  It wasn't obvious in the spec.  Is it there and I'm just missing it?  If it's not there, is it still useful to have?

https://github.com/tetrascience/ts-task-script-utils/pull/2#discussion_r770882278
> This table is for 2 digit years as far as I understand. A 4 digit year should never depend on the "year-first" config value.
> 
> Maybe we need a separate table for 4 digit years, but I think it would be a subset of this table (i.e. everywhere we have `year-first` `True` in the table corresponds to a 4-digit-year-format of `YYYY-XX-XX`; anywhere with `year-first` `False` corresponds to `XX-XX-YYYY`; anywhere with `year-first` `None` isn't included because a 4 digit year is unambiguous)

We have no table for 4-digit years.  Do we still want one?

https://github.com/tetrascience/ts-task-script-utils/pull/2#discussion_r770229088
> We should explicitly say if `year_first` is `True`, then we assume the format will be YMD because there is no YDM format in any country. It's a very safe assumption and backed by https://en.wikipedia.org/wiki/Date_format_by_country#

We never say why YDM is not a valid format when `year_first` is True.  Is it still important to do so?  At one time you thought it was because you even [mention it again](https://github.com/tetrascience/ts-task-script-utils/pull/2#discussion_r770230574).

README.md
```diff
@@ -6,6 +6,12 @@ Utility functions for Tetra Task Scripts
 
 - [Installation](#installation)
 - [Usage](#usage)
+- [Datetime Parser](#datetime-parser)
+  - [Usage](#usage-1)
+  - [DatetimeConfig](#datetimeconfig)
+  - [Datetime format resolution matrix](#datetime-format-resolution-matrix)
+  - [Format Tokens](#format-tokens)
+  - [Limitations](#limitations)
```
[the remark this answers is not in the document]
setiat: fixed.

__tests__/integration/test_datetime_parser.py
```diff
@@ -0,0 +1,127 @@
...
+    datetime_config_dict = {"tz_dict": tz_dict}
+
+    datetime_config = DatetimeConfig(**datetime_config_dict)
+    try:
+        parsed_datetime = parse(
+            input,
+            formats_list,
+            datetime_config,
+        )
+    except Exception as e:
```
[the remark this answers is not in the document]
setiat: fixed.

task_script_utils/datetime_parser/README.md
```diff
@@ -0,0 +1,385 @@
...
+- `datetime_raw_str: str`: raw datetime string to be parsed.
+- `formats_list: Sequence[str] (optional)`: You can optionally pass a list of formats to try to parse datetime string. If `datetime_raw_str` doesn't matches with any format, the datetime parser will still try to parse `datetime_raw_str` with other methods such as using regex and trying long datetime format
+- `config: DatetimeConfig (optional)`: You also have an options to pass `DatetimeConfig`. It provides complementary information on how to mark parsed digits as day, month or year and also provide an options to handle abbreviated time zones and fold for parsing ambiguous timestamps during daylight saving transitions. Ideally, DatetimeConfig should be constructed from pipeline configuration passed to task scripts
+
+Output:
+
+- A `TSDatetime` object.
+
+![Parse Flow Diagram](./parse-white.png)
+![Short date format Resolution for two digit date parts](./short-date-format-resolution.png)
```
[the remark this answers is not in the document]
setiat: fixed

setup.py
```diff
@@ -3,12 +3,16 @@
 from setuptools import setup, find_packages
 
 NAME = "ts-task-script-utils"
-VERSION = "1.0.1"
+VERSION = "1.2.0"
```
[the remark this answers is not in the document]
setiat: I wanted to go for 1.1.0 but then there is a ticket DE-2985 that says release v1.2.0.
Should I change it to 1.1.0?

__tests__/integration/test_datetime_parser.py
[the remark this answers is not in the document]
setiat: The explicit datetime config, make it un-parsable in this case.
But if user does not specify any datetime config or explicitly set year_first and day_first to None, it will be parsed as `2021-11-14T04:03:00-05:00`

task_script_utils/datetime_parser/datetime_info.py
```diff
@@ -496,7 +496,7 @@ def long_date_format(self) -> str:
                 f"YYYY"
             )
         else:
-            date_fmt = f"{self.token_month} " f"{self.token_day} " f"YYYY"
+            date_fmt = date_fmt = f"{self.token_month} {self.token_day} YYYY"
```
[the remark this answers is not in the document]
setiat: fixed

setup.py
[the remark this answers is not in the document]
mlipschultz: Yes.  I don't know why I wrote 1.2.0 in the ticket, sorry!

__tests__/integration/test_datetime_parser.py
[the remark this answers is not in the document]
mlipschultz: According to the flow chart, if we make it to "Try Regex parsing using DatetimeInfo" (which we do for this case) it says "If config is passed to parse(), it will be used to resolve ambiguity when inferring date".  In this case, I don't think there is any ambiguity so it shouldn't need to consult config.

setup.py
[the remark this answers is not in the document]
setiat: Changed it v1.0.0

__tests__/integration/test_datetime_parser.py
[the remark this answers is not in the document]
setiat: The readme and flow chart have now been updated to indicated that if `DatetimeConfig` is present, it has higher precedence.

task_script_utils/datetime_parser/README.md
```diff
@@ -0,0 +1,385 @@
...
+
+## Usage
+
+`from task_script_utils.datetime_parser import parse`
+
+Input:
+
+- `datetime_raw_str: str`: raw datetime string to be parsed.
+- `formats_list: Sequence[str] (optional)`: You can optionally pass a list of formats to try to parse datetime string. If `datetime_raw_str` doesn't matches with any format, the datetime parser will still try to parse `datetime_raw_str` with other methods such as using regex and trying long datetime format
+- `config: DatetimeConfig (optional)`: You also have an options to pass `DatetimeConfig`. It provides complementary information on how to mark parsed digits as day, month or year and also provide an options to handle abbreviated time zones and fold for parsing ambiguous timestamps during daylight saving transitions. Ideally, DatetimeConfig should be constructed from pipeline configuration passed to task scripts
```
mlipschultz, line 22: [minor]
```suggestion
- `config: DatetimeConfig (optional)`: You also have an options to pass `DatetimeConfig`. It provides complementary information on how to mark parsed digits as day, month or year and also provide options to handle abbreviated time zones and fold for parsing ambiguous timestamps during daylight saving transitions. Ideally, DatetimeConfig should be constructed from pipeline configuration passed to task scripts
```

mlipschultz (approved):
I have one minor comment regarding wording, but everything looks good to merge!

wkopen (approved):
@mlipschultz I didn't look too close. But many of my comments have been addressed in another PR I believe. Thus approving this.

PR #3, opened 2021-11-10
DE-2591: datetime config

setiat:
I have renamed `PipelineConfig` to `DatetimeConfig` as the name makes more sense for this use-case.
Created a separate PR, just in case it is a bad idea and we all don't agree.


robjhornby-ts:
I agree `DatetimeConfig` is a better name than `PipelineConfig` because `PipelineConfig` already has a specific meaning to a pipeline/protocol; this utility should be decoupled from that meaning (the configuration might come from a pipeline configuration, but in general it can come from any source)

PR #4, opened 2021-11-11
DE-2591: subseconds handling

setiat:
- ISO datetime result must contain same digit for sub seconds as in input string.
- This still fails for cases where n(digits) > 6. `Pendulum` allows parsing up to 6 digits
- if datetime is parsed using a format from formats_list, this doesn't apply and default `.SSS` will be used


task_script_utils/datetime_parser/convert_to_iso_format.py
```diff
@@ -23,20 +26,36 @@ def convert_to_ts_iso8601(
...
+    default_format = "YYYY-MM-DDTHH:mm:ss"
+
+    if datetime_info is None:
+        return f"{default_format}.SSS"
+
+    if datetime_info.fractional_seconds is None:
+        return f"{default_format}.SSS"
+
+    subseconds = datetime_info.fractional_seconds
+    subseconds_token = "S"* len(subseconds)
```
mlipschultz, line 59:
```suggestion
    subseconds_token = "S" * len(subseconds)
```

mlipschultz (commented):
This all looks good to me.  The only thing I'm not comfortable with is:
> if datetime is parsed using a format from formats_list, this doesn't apply and default .SSS will be used
I'd think we'd always want to preserve the precision of the measurement (or lack of it).  Was this a decision somewhere I missed?

robjhornby-ts:
> This all looks good to me. The only thing I'm not comfortable with is:
> 
> > if datetime is parsed using a format from formats_list, this doesn't apply and default .SSS will be used
> 
> I'd think we'd always want to preserve the precision of the measurement (or lack of it).  Was this a decision somewhere I missed?

Agreed, ideally we'd preserve subsecond precision even when using `formats_list`, it would be nice if there was a generic token for "subseconds with any amount of precision" instead of pendulum's hardcoded `"S", "SS", "SSS"...`. Maybe a simple interface would be for us to always catch any of those `"S"` tokens in the formats_list input and instead use logic for any level of precision. Would that be straightforward @tariksetia ? We'd be using a slightly different syntax from Pendulum then, but as long as we document the difference I think it's a more useful way of doing it

setiat:
- We can use a generic token say `X` . But then, for each format in the `formats_list`, there would be 6 option to try `.S, .SS ... .SSSSSS`. This is ok..but the main problem is sometimes .SS can parse the timestamps where fractional seconds actually corresponds to format containing .SSS. This makes the `microseconds` in the result wrong. I will see how this works in pendulum.

```python
#Incorrect Way
>>> import pendulum as p
>>> test = p.from_format("13:00:00.010", "HH:mm:ss.SS")
>>> test
DateTime(2021, 11, 15, 13, 0, 0, 100000, tzinfo=Timezone('UTC'))
>>> test.format("SS")
'10'
>>> test.format("SSS")
'100'
#Correct Case
>>> test_new = p.from_format("13:00:00.010", "HH:mm:ss.SSS")
>>> test_new
DateTime(2021, 11, 15, 13, 0, 0, 10000, tzinfo=Timezone('UTC'))
>>> test_new.format("SS")
'01'
>>> test_new.format("SSS")
'010'
```

- This also makes parse() useless in a sense that the parsed `datetime` object may have wrong fractional seconds(as in example above).
  - This can be corrected by returning a tuple or new object containing `datetime` object with (may be) wrong fractional seconds and an `int` value representing fractional value in raw string.
  - This tuple can be used to build tetrascience compatible ISO datetime string.
  - But parse will now be a part of internal api and should be exposed to developer.

- Another issue is the placement of fractional seconds in datetime format... eg hh:mm:ss.SSS is easy to parse but if `SSS` can be places any where in the format string, parsing out the actual value of fractional seconds would be hard.

There could be an easier way, which I am not able to think of right now... So, Any help/suggestion is highly welcomed and deeply appreciated :)

mlipschultz:
I was hoping we could just always do `SSSSSS` and pendulum would just say "the only digits I see are `010`, so I'll just extract it as `0.010` in the match", but it doesn't:
```python
>>> import pendulum as p
>>> test_new = p.from_format("13:00:00.010", "HH:mm:ss.SSSSSS")
>>> test_new
DateTime(2021, 11, 15, 13, 0, 0, 10, tzinfo=Timezone('UTC'))
>>> test_new.format("SSS")
'000'
```

So it looks like pendulum will start at the right most part of a match to `S*` and work its way left to capture as many sub-second digits as it can, padding with extra zeros on the left as necessary.  So in the example, `0.010` became `0.000010`.  I wonder what the reasoning was for this implementation.

As an alternative to your idea, how about this: We do the pattern matching ourselves using string substitution and regular expressions.  We adhere to the pendulum format codes (except for subsecond, where `S` means "match one or more sub-second digits").  Here's a simple implementation example:
```python
>>> import re
>>> from decimal import Decimal
>>> format = "HH:mm:ss.S"
>>> re_format = format.replace('HH', r'(?P<hour>\d\d)').replace('mm', r'(?P<minute>\d\d)').replace('ss', r'(?P<second>\d\d)').replace('S', r'(?P<subsecond>\d+)')
>>> match = re.fullmatch(re_format, "13:00:00.010")
>>> hour = int(match.group('hour'))
>>> minute = int(match.group('minute'))
>>> second = int(match.group('second'))
>>> subsecond = Decimal('0.' + match.group('subsecond'))
>>> (hour, minute, second, subsecond)
(13, 0, 0, Decimal('0.010'))
```
To actually use this, we'd need to:
1. improve `format.replace`
2. check the captured groups to make sure they're valid values (e.g. `hour` can't be `99`)
2. replace the `(hour, minute, second, subsecond)` with an actual date time object that can handle `subsecond` in a way that we need.

I also know there's a danger in having to implement this ourselves (more code to maintain, less tested than pendulum, ...).  However, @setiat did ask for other ideas :).  I'm not saying this is a better idea, just a different one.

mlipschultz (changes requested):

__tests__/unit/test_convert_to_iso.py
```diff
@@ -5,23 +5,55 @@
...
+    ("2021-11-07T04:30:00.1 America/New_York", "2021-11-07T09:30:00.1Z"),
+    ("2021-11-07T04:30:00.12 America/New_York", "2021-11-07T09:30:00.12Z"),
+    ("2021-11-07T04:30:00.123 America/New_York", "2021-11-07T09:30:00.123Z"),
+    ("2021-11-07T04:30:00.01230 America/New_York", "2021-11-07T09:30:00.01230Z"),
+    ("2021-11-07T04:30:00.123456 America/New_York","2021-11-07T09:30:00.123456Z"),
+    ("2021-11-07T04:30:00.00123456 America/New_York","2021-11-07T09:30:00.00123456Z"),
+    ("2021-11-07T04:30:00.01 America/New_York","2021-11-07T09:30:00.01Z"),
+    ("2021-11-07T04:30:00.010 America/New_York","2021-11-07T09:30:00.010Z"),
+    # Error Case as length of subseconds > 7
+    ("2021-11-07T04:30:00.1234567 America/New_York", '2021-11-07T09:30:00.1234567Z'),
```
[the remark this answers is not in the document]
setiat: Fixed... this not an error case anymore. Parser will now be handling decimal digits of arbitrary lengths

task_script_utils/datetime_parser/ts_datetime.py
```diff
@@ -0,0 +1,72 @@
...
+            if self._subseconds is not None:
+                iso_8601 = f"{iso_8601}.{self._subseconds}"
+
+        return iso_8601
+
+    @property
+    def iso_format(self):
+        iso_str = self._datetime.format("YYYY-MM-DDTHH:mm:ss")
+        if self._subseconds:
+            iso_str += f".{str(self._subseconds)[:6]}"
```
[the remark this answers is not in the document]
setiat: I have changed it... it will not truncated digits anymore.
Will change it later if rounding is required.

task_script_utils/datetime_parser/ts_datetime.py
```diff
@@ -0,0 +1,72 @@
...
+            iso_8601 = f"{iso_8601}Z"
+        else:
+            iso_8601 = self._datetime.format(minimal_format)
+            if self._subseconds is not None:
+                iso_8601 = f"{iso_8601}.{self._subseconds}"
+
+        return iso_8601
+
+    @property
+    def iso_format(self):
```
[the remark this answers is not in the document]
setiat: added.

task_script_utils/datetime_parser/datetime_info.py
```diff
@@ -431,7 +427,8 @@ def dtstamp(self):
             dt_str = f"{self.day}-{self.month}-{self.year}"
             dt_str += f" {self.hour}:{self.minutes}:{self.seconds}"
             if self.fractional_seconds:
-                dt_str += f".{self.fractional_seconds}"
+                # Pendulum support upto 6 fractional seconds
+                dt_str += f".{self.fractional_seconds[:6]}"
```
mlipschultz, line 431: Why truncate to 6 fractional seconds?  I thought we handled arbitrary length now,

setiat: Python `datetime` object allow unto 6 digit fractional seconds. These are actually `microseconds %f`
```python
>>> from datetime import datetime
>>> dt_string = "12/11/2018 09:15:32"
>>> dt_object1 = datetime.strptime(dt_string, "%d/%m/%Y %H:%M:%S")
>>> dt_string = "12/11/2018 09:15:32.1234567"
>>> dt_object1 = datetime.strptime(dt_string, "%d/%m/%Y %H:%M:%S.%f")
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Library/Frameworks/Python.framework/Versions/3.9/lib/python3.9/_strptime.py", line 568, in _strptime_datetime
    tt, fraction, gmtoff_fraction = _strptime(data_string, format)
  File "/Library/Frameworks/Python.framework/Versions/3.9/lib/python3.9/_strptime.py", line 352, in _strptime
    raise ValueError("unconverted data remains: %s" %
ValueError: unconverted data remains: 7
>>> 
```

The parse function returns, `TSDatetime`. It has two attributes:
1. `TSDatetime._datetime`: python `datetime` object with unto 6 fractional seconds
2. `TSDatetime._subseconds: int`: integral value of decimal value

these two attributes can be used to build `ids_format` and `ts_format`
But if developer is looking for python datetime object `TSDatetime.datetime` will return it with truncated decimals.

mlipschultz: But there's nothing in this method's name or description that indicate it's only to be used for python datetime.

setiat: My Bad. Fixed it by adding inline comments

mlipschultz (approved):
I'm not sure if you want @robjhornby-ts's review on this too, so I'll leave it up to you whether this is merged or not.

PR #7, opened 2021-12-17
DE-3026: Black formatting check

mlipschultz:


.github/workflows/code_quality.yml
```diff
@@ -0,0 +1,18 @@
...
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+    - name: Set up Python
+      uses: actions/setup-python@v2
+      with:
+        python-version: '3.x'
+    - name: Black
+      run: |
+        python -m pip install black
```
VladimirShteyn, line 17: I recommend pinning the version of `black` because formatting of edge cases changes slightly from version to version. If I recall correctly `v20.*` had a lotta bugs in it, but `v21.b12` seems to be working fine for me now. DI team should be notified that their Task Scripts' dependencies should be compatible with those of `black` `v21.b12`.

.github/workflows/code_quality.yml
```diff
@@ -0,0 +1,18 @@
+name: Code Quality
+
+on: [push]
+
+jobs:
+  build:
+
+    runs-on: ubuntu-latest
```
VladimirShteyn, line 8: [out of ticket scope] [vlad edification] Are there other options we should consider? For example, in the future we could have multiple steps -- possibly 10+ -- in this config file and CI would go faster if all CI were executed in one pre-built environment.

.github/workflows/code_quality.yml
```diff
@@ -0,0 +1,18 @@
...
+    steps:
+    - uses: actions/checkout@v2
+    - name: Set up Python
+      uses: actions/setup-python@v2
+      with:
+        python-version: '3.x'
```
VladimirShteyn, line 14: I like the flexibility here.

VladimirShteyn (approved):

.github/workflows/code_quality.yml
mlipschultz: Great idea!

.github/workflows/code_quality.yml
mlipschultz: I'm not sure if we can use a custom image, but it looked like we have other options.

On this topic, I like that you can create more pipelines and give them whatever names you want.  So if we just wanted a code quality pipeline, we could have it.  Or if we wanted to have alerts that wouldn't block the build, that might be possible.

mlipschultz: Also, there's a marketplace for github actions: https://github.com/marketplace?type=actions

PR #13, opened 2022-02-16
DE-2999 separate long and short date parsing

setiat:
Creating this separate PR for moving  unrelated long_datetime parsing methods from DatetimeInfo to their own module


task_script_utils/datetime_parser/long_datetime_utils.py
```diff
@@ -0,0 +1,71 @@
...
+    ["m", "mm"],
+    ["s", "ss"],
+]
+
+
+def get_time_formats_for_long_date(fractional_seconds):
+    def map_am_pm(time_format):
+        return time_format if time_format.startswith("H") else time_format + " A"
+
+    time_formats = [":".join(tokens) for tokens in product(*TIME_PARTS)]
```
mlipschultz, line 20: Will there always be seconds in a long date format?

task_script_utils/datetime_parser/long_datetime_utils.py
```diff
@@ -0,0 +1,71 @@
...
+    if fractional_seconds:
+        token = "SSSSSS"
+        time_formats = map(lambda x: [x, f"{x}.{token}"], time_formats)
+
+    time_formats = flatten(time_formats)
+    time_formats = map(lambda x: map_am_pm(x), time_formats)
+    time_formats = map(
+        lambda x: [x, x + " Z", x + " z", x + " ZZ", x + " Z z", x + " ZZ z"],
```
mlipschultz, line 28: Any reason `" zz"`, `" Z zz"`, and `" ZZ zz"` aren't included?

task_script_utils/datetime_parser/utils.py
```diff
@@ -1,6 +1,7 @@
 import re
 import datetime as dt
-from typing import Sequence
+from sqlite3 import Date
```
mlipschultz, line 3: Some of these imports are unused

task_script_utils/datetime_parser/parser.py
```diff
@@ -58,9 +59,11 @@ def parse(
 
     # Use long date formats
     if not parsed_datetime:
+        datetime_info._parse_long_date_formats()
```
mlipschultz, line 62: You're accessing a protected method outside of the class.  Moving all long date code out of DateTimeInfo would resolve this problem

task_script_utils/datetime_parser/datetime_info.py
```diff
@@ -473,48 +470,6 @@ def datetime(self):
...
-
-        Raises:
-            InvalidDateError: Raised if parsing fails to
-            detect tokens for month or day
-
-        Returns:
-            str: Return Date format built using pendulum
-            formatting tokens
-        """
-        self._parse_long_date_formats()
```
mlipschultz, line 490: I'm not seeing this method removed from DateTimeInfo.  I'd also like to see:
1. `_parse_long_date_formats` moved out of this class
2. `_get_matchers_list` updated to only have the `long_date_formats=False` matchers
3. `_match_day_of_week_token`, `_match_month_token`, and `_match_day_token` moved out of this class

I'm seeing `DateTimeInfo` being a class that parses a string and extracts the date and time information.  The long date code in this class isn't doing that.  It's just identifying what tokens to use for parsing and that parsing was being done outside of this class already.  I know I originally suggested moving all long date code out of the class, but I think what's more correct is to put all long date code in one place (either all in this class or all out of this class).  Because of how separate the flows are between short and long parsing, I think it makes sense to separate the parsing.  However, I can see an advantage to having the final object look similar.  Perhaps it's best to have:
1. a `DateTimeInfo` class that's just responsible for:
    1. holding date and time information (i.e. `am_or_pm`, `iana_tz`, ..., `fractional_seconds`)
    2. all of the `@property` methods currently in `DateTimeInfo`
2. a `ShortDateTimeInfo` class that inherits from `DateTimeInfo` that's responsible for:
    1. containing all the methods for parsing a string using the short date time methods and setting all the date and time variables from the parent class
3. a `LongDateTimeInfo` class that inherits from `DateTimeInfo` that's responsible for:
    1. containing all the methods for parsing a string using the long date time methods and setting all the date and time variables from the parent class

You'll still need code to determine whether it's a long or short datetime, but you already have that so it's just a matter of updating the code.

mlipschultz (changes requested):

task_script_utils/datetime_parser/long_datetime_utils.py
setiat: Again long datetime parser was written in the context of empower agent.
But I still forgot `x + " z ZZ",
I have added all of them

task_script_utils/datetime_parser/utils.py
setiat: removed.

task_script_utils/datetime_parser/parser.py
setiat: fixed.

task_script_utils/datetime_parser/datetime_info.py
setiat: @mlipschultz Made the required changes. 
Doc strings and flowchart are still pending. Will wait till new changes are approved.

mlipschultz (changes requested):
Thanks for separating the logic into separate classes!

In [__tests__/integration/test_datetime_parser.py](https://github.com/tetrascience/ts-task-script-utils/pull/13/files#diff-896bdb7560d9b4a7b0ec65de321fd56633f238de5eb5a8f6b38b8a1b1062cc2a), lines 46 and 47 are duplicates of lines 39 and 40.

task_script_utils/datetime_parser/long_datetime_utils.py
mlipschultz: But what about other agents or sources for dates?  I can see that there exists at least one source for long dates that includes seconds, but is that true for all sources?

task_script_utils/datetime_parser/utils.py
```diff
@@ -116,3 +117,31 @@ def replace_z_with_offset(datetime_str: str) -> str:
...
+    )
+    if datetime_str_with_no_abbreviated_tz != datetime_str:
+        # It means datetime_str did contain abbreviated_tz and we
+        # have replaced it with its utc_offset value from tz_dict.
+        # Now if the format in formats_list contains "zz", replace
+        # it with "Z". This is because no library parses abbreviated tz
+        # due to its ambiguous nature
+        formats_with_no_zz = replace_zz_with_Z(formats)
+    else:
+        formats_with_no_zz = formats
```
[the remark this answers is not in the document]
setiat: fixed

task_script_utils/datetime_parser/utils.py
```diff
@@ -116,3 +117,31 @@ def replace_z_with_offset(datetime_str: str) -> str:
...
+
+    for format_ in formats_with_no_zz:
+        try:
+            parsed = from_pendulum_format(
+                datetime_str_with_no_abbreviated_tz, format_, tz=None
+            )
+            return parsed, format_
+        except Exception as e:
```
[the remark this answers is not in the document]
setiat: fixed.

Catch -> ValueError and re.error

task_script_utils/datetime_parser/utils.py
```diff
@@ -116,3 +117,31 @@ def replace_z_with_offset(datetime_str: str) -> str:
     12-12-12T14:53:00 Z -> 12-12-12T14:53:00 +00:00
     """
     return re.sub(r"(?<=\d|\s)Z(?=\s|$)", "+00:00", datetime_str)
+
+
+def parse_with_formats(datetime_str: str, datetime_config, formats: Sequence[str] = ()):
```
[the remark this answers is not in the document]
setiat: added.

task_script_utils/datetime_parser/parser.py
```diff
@@ -43,26 +37,19 @@ def parse(
     datetime_str = replace_z_with_offset(datetime_raw_str)
     # Parse Using formats list
     if formats_list:
-        parsed_datetime, matched_format = _parse_with_formats(
+        parsed_datetime, matched_format = parse_with_formats(
             datetime_str, datetime_config=config, formats=formats_list
```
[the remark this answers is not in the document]
setiat: No particular reason.
Is this a bad habit?

task_script_utils/datetime_parser/datetime_info.py
```diff
@@ -68,39 +52,343 @@ def _parse(self, matchers):
                     # remove it from the _matcher list
                     _matchers.remove(func)
 
+    def _pre_process_datetime_string(self):
+        """This method is used to pre-process the input string
+        if required, before the parsing starts. It performs following
+        pre-processing:
+        1. Repace a character if it is sandwiched between two integer
```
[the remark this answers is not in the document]
setiat: Fixed docstring.
I would like to rename it to `"remove 'T' between two digits"` as we can not ignore the possibility of getting `11:12:23A2018-13-09`

Waiting for @robjhornby-ts 's view on this.

task_script_utils/datetime_parser/datetime_info.py
```diff
@@ -68,39 +52,343 @@ def _parse(self, matchers):
...
+                dt_str += f" {self.iana_tz}"
+
+            return dt_str
+
+        return None
+
+    @property
+    def datetime_format(self):
+        """Build datetime format using pendulum tokens.
+        This property along with `dtstamp` property can be used
```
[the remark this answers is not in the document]
setiat: fixed

__tests__/integration/test_datetime_parser.py
```diff
@@ -195,18 +194,6 @@ def test_parse_with_formats(input, expected):
         assert parsed_datetime.isoformat() == expected
 
 
-@pytest.mark.parametrize(
-    "input, expected", format_list_with_no_tz_dict_test_cases.items()
```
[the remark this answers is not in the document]
setiat: removed.

task_script_utils/datetime_parser/datetime_info.py
```diff
@@ -68,39 +52,343 @@ def _parse(self, matchers):
...
+            self.datetime_stamp, self.datetime_format, tz=None
+        )
+        ts_datetime = TSDatetime(
+            datetime_=datetime_, subseconds=self.fractional_seconds
+        )
+        return ts_datetime
+
+    @datetime.setter
+    def datetime(self, ts_datetime: TSDatetime):
+        """This setter is mainly used by LongDateTimeInfo
```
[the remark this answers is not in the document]
setiat: `datetime` is a property.  This is why I went for `datetime.setter`

task_script_utils/datetime_parser/long_datetime_utils.py
setiat: Long Datetime parser was written from the perspective empower's date-time format.
We still can have other sources... But we also provide a list of possible formats as format list

I am ok with whatever is decided here. 

It will just double the number of formats in the resultant format_list

mlipschultz: Since this is a utility meant to be used across task scripts, I'd rather have a general-purpose solution.  @robjhornby-ts, what are your thoughts on this since you have a better understanding than me of task script needs

task_script_utils/datetime_parser/parser.py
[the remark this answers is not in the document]
mlipschultz: I think it was ok, it just made me wonder if I was missing something -- was there some parameter that wasn't being specified?

task_script_utils/datetime_parser/datetime_info.py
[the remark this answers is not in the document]
mlipschultz: Thanks!  Looks like you missed the 'l' in "Replace" and the 's' in "integers".

I'll reach out to Rob on slack for the question.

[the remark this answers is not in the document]
robjhornby-ts: I think it makes sense to have this just remove `T` for now. If we come across another case that needs it we can update it, that seems like a safe option.

It looks like there's currently only 1 preprocessing step. To me it would make more sense to:
- only store the `date_time_raw` string
- remove `date_time_processed`
- put the `T` removing logic into a tokenization function (maybe a "_tokenize_datetime_string" function) which can be used here instead of using `.split()`: https://github.com/tetrascience/ts-task-script-utils/blob/18fcc19af9590d6354690158ffda58952b9fb564/task_script_utils/datetime_parser/datetime_info.py#L44?

The reason is that this step of replacing a letter with a space feels like it's an intermediate part of something else, so storing this intermediate result `date_time_processed` might not be that useful for other parts of the code. Storing the raw string and having 1 function which can tokenize it into tokens which are ready to use with `matchers` feels nicely encapsulated. Then if we ever need to add other rules about how we tokenize a raw datetime string before it's passed to `matchers`, that can all happen in the `_tokenize_datetime_string` function

task_script_utils/datetime_parser/long_datetime_utils.py
robjhornby-ts: I agree we'd ideally make it general-purpose but I also haven't seen many long datetime formats without `seconds` yet so I think it's ok to require seconds for the immediate cases. How easy would it be to make the range of long datetime formats configurable? (Probably not easy unless you'd already anticipated this...)

How many other long date formats are we not covering? Maybe it's not possible to cover them all without some custom work per-task-script (and we can work out how to make that custom work as easy as possible, e.g. some kind of helper function for creating a list of long date formats, then that list of formats can be input to the usual format-list parser)

task_script_utils/datetime_parser/datetime_info.py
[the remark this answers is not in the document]
mlipschultz: I was asking why use this setter method here: https://github.com/tetrascience/ts-task-script-utils/blob/18fcc19af9590d6354690158ffda58952b9fb564/task_script_utils/datetime_parser/datetime_info.py#L246
when you could directly set day, month, ... there instead.  Looking back, my comment wasn't very clear, sorry!

I think what I'd like to see is a change to the docsting.  It doesn't need to talk about `LongDateTimeInfo`.  Unfortunately, I can't think of what the docstring should say since it seems self-explanatory to me.  Maybe it's not necessary?

mlipschultz (commented):
Thanks, looks like almost all my concerns are addressed!  I reached out to Rob for some clarification on requirements.  I think my remaining requests are all minor things due to miscommunication.

task_script_utils/datetime_parser/long_datetime_utils.py
mlipschultz: Based on Rob's response, it sounds like:
1. For this PR, we keep what you have
2. We can create a ticket for the helper function that'll create long datetime formats

@robjhornby-ts, to answer your question about how many long date formats we aren't covering, I think we cover all other cases.  Tarik did a thorough job of creating every combination of possibilities: https://github.com/tetrascience/ts-task-script-utils/blob/18fcc19af9590d6354690158ffda58952b9fb564/task_script_utils/datetime_parser/datetime_info.py#L330

mlipschultz:
It looks like you missed this (and I forgot about it in my previous review):

> In [\_\_tests\_\_/integration/test_datetime_parser.py](https://github.com/tetrascience/ts-task-script-utils/pull/13/files#diff-896bdb7560d9b4a7b0ec65de321fd56633f238de5eb5a8f6b38b8a1b1062cc2a), lines 46 and 47 are duplicates of lines 39 and 40.

task_script_utils/datetime_parser/datetime_info.py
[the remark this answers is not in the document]
setiat: - added `_tokenize_datetime_string ` and `_remove_T_between_two_digits`
- removed `date_time_processed ` and `_preprocess_Datetime_str`

task_script_utils/datetime_parser/datetime_info.py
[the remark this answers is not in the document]
setiat: Removed the docstrings for datetime.setter

setiat:
> It looks like you missed this (and I forgot about it in my previous review):
> 
> > In [__tests__/integration/test_datetime_parser.py](https://github.com/tetrascience/ts-task-script-utils/pull/13/files#diff-896bdb7560d9b4a7b0ec65de321fd56633f238de5eb5a8f6b38b8a1b1062cc2a), lines 46 and 47 are duplicates of lines 39 and 40.

Removed duplicate data

task_script_utils/datetime_parser/datetime_info.py
```diff
@@ -52,24 +51,24 @@ def _parse(self, matchers):
...
 
-    def _pre_process_datetime_string(self):
+    def _tokenize_datetime_string(self):
         """This method is used to pre-process the input string
-        if required, before the parsing starts. It performs following
-        pre-processing:
-        1. Repace a letter if it is sandwiched between two integer
+        and return token list splitted by whitespace
+        It performs following pre-processing:
+        1. Repace the letter T if it is sandwiched between two digits
```
mlipschultz, line 58:
```suggestion
        1. Replace the letter T if it is sandwiched between two digits
```

mlipschultz (commented):
Just one minor typo to fix, then it's ready to merge!

task_script_utils/datetime_parser/datetime_info.py
setiat: fixed.

PR #19, opened 2022-03-29
FIX: Change travis file to specify ubuntu version

abclay:
We need to pin the version of ubuntu that travis uses to run, otherwise we get errors.  Not sure exactly why.


mlipschultz (approved):
So this is why I started seeing `dist: focal` in some PRs

PR #21, opened 2022-04-15
DE-3599: Add parsing function stubs and tests

abclay:
## JIRA ticket
https://tetrascience.atlassian.net/browse/DE-3599

## Description

I have added all the test cases from [Confluence](https://tetrascience.atlassian.net/wiki/spaces/DE/pages/3230400808/Unit+Test+Cases+for+Parsing+Primitives#Parse-string-to-boolean)
I have also added stub functions to be tested.  I wrote very basic implementations to check my tests and left them in. When we do the next ticket to implement these methods, we can use the reference implementations as a baseline.

## Automated testing

- [x] Unit tests
- [ ] Integration tests
- [ ] API tests
- [ ] End-To-End tests
- [ ] N/A (please clarify)

### DIFF
--- a/__tests__/unit/parse_functions/test_to_boolean.py
+++ b/__tests__/unit/parse_functions/test_to_boolean.py
@@ -1,0 +1,217 @@
+"""
+Test `to_boolean()` function in parse.py
+"""
+import pytest
+from task_script_utils import parse
+
+
+@pytest.mark.parametrize("value", ["True", "False", "Something else"])
+@pytest.mark.parametrize("case_sensitive", ["True", "False"])
+def test_to_boolean_empty_sets(value, case_sensitive):
+    """Test Empty Sets"""
+    # Arrange
+
+    # Act
+
+    # Assert
+    with pytest.raises(Exception):
+        parse.to_boolean(value, {}, {}, case_sensitive=case_sensitive)
+
+
+@pytest.mark.parametrize("value", ["True", "False", "Something else"])
+@pytest.mark.parametrize("case_sensitive", ["True", "False"])
+def test_to_boolean_empty_true_set(value, case_sensitive):
+    """Test Empty True Set"""
+    # Arrange
+
+    # Act
+
+    # Assert
+    with pytest.raises(Exception):
+        parse.to_boolean(value, {}, {"False"}, case_sensitive=case_sensitive)
+
+
+@pytest.mark.parametrize("value", ["True", "False", "Something else"])
+@pytest.mark.parametrize("case_sensitive", ["True", "False"])
+def test_to_boolean_empty_false_set(value, case_sensitive):
+    """Test Empty False Set"""
+    # Arrange
+
+    # Act
+
+    # Assert
+    with pytest.raises(Exception):
+        parse.to_boolean(value, {"true", "1", "yes"}, {}, case_sensitive=case_sensitive)
+
+
+@pytest.mark.parametrize("value", ["True", "1", "yes"])
+def test_to_boolean_case_sensitive_string_is_in_true_set(value):
+    """Test Case sensitive string is in True Set"""
+    # Arrange
+
+    case_sensitive = True
+    # Act
+
+    actual = parse.to_boolean(
+        value, {"True", "1", "yes"}, {"False", "0", "no"}, case_sensitive=case_sensitive
+    )
+
+    # Assert
+    assert isinstance(actual, bool)
+    assert actual is True
+
+
+@pytest.mark.parametrize(
+    "value", ["yes", "Yes", "yEs", "yeS", "YEs", "YeS", "yES", "YES"]
+)
+def test_to_boolean_case_insensitive_string_is_in_true_set(value):
+    """Test Case insensitive string is in True Set"""
+    # Arrange
+
+    case_sensitive = False
+    # Act
+
+    actual = parse.to_boolean(
+        value, {"True", "1", "yes"}, {"False", "0", "no"}, case_sensitive=case_sensitive
+    )
+
+    # Assert
+    assert isinstance(actual, bool)
+    assert actual is True
+
+
+@pytest.mark.parametrize("value", ["False", "0", "no"])
+def test_to_boolean_case_sensitive_string_is_in_false_set(value):
+    """Test Case sensitive string is in False Set"""
+    # Arrange
+
+    case_sensitive = True
+    # Act
+
+    actual = parse.to_boolean(
+        value, {"True", "1", "yes"}, {"False", "0", "no"}, case_sensitive=case_sensitive
+    )
+
+    # Assert
+    assert isinstance(actual, bool)
+    assert actual is False
+
+
+@pytest.mark.parametrize("value", ["no", "No", "nO", "NO"])
+def test_to_boolean_case_insensitive_string_is_in_false_set(value):
+    """Test Case insensitive string is in False Set"""
+    # Arrange
+
+    case_sensitive = False
+    # Act
+
+    actual = parse.to_boolean(
+        value, {"True", "1", "yes"}, {"False", "0", "no"}, case_sensitive=case_sensitive
+    )
+
+    # Assert
+    assert isinstance(actual, bool)
+    assert actual is False
+
+
+@pytest.mark.parametrize(
+    "value",
+    ["TRUE", "FALSE", "Yes", "yEs", "yeS", "YEs", "YeS", 'yES""YES', "No", "nO", "NO"],
+)
+def test_to_boolean_case_sensitive_string_is_not_in_either_set__but_looks_like_it_should_be(
+    value,
+):
+    """Test Case sensitive string is not in either set, but looks like it should be"""
+    # Arrange
+
+    case_sensitive = True
+    # Act
+
+    actual = parse.to_boolean(
+        value, {"True", "1", "yes"}, {"False", "0", "no"}, case_sensitive=case_sensitive
+    )
+
+    # Assert
+    assert actual is None
+
+
+@pytest.mark.parametrize("value", ["    True", " 1", "\\t   \\t yes"])
+def test_to_boolean_leading_whitespace__true_(value):
+    """Test Leading whitespace (true)"""
+    # Arrange
+
+    case_sensitive = True
+    # Act
+
+    actual = parse.to_boolean(
+        value, {"True", "1", "yes"}, {"False", "0", "no"}, case_sensitive=case_sensitive
+    )
+
+    # Assert
+    assert isinstance(actual, bool)
+    assert actual is True
+
+
+@pytest.mark.parametrize("value", ["True   ", "1\\t", "yes     \\t\\t"])
+def test_to_boolean_trailing_whitespace__true_(value):
+    """Test Trailing whitespace (true)"""
+    # Arrange
+
+    case_sensitive = True
+    # Act
+
+    actual = parse.to_boolean(
+        value, {"True", "1", "yes"}, {"False", "0", "no"}, case_sensitive=case_sensitive
+    )
+
+    # Assert
+    assert isinstance(actual, bool)
+    assert actual is True
+
+
+@pytest.mark.parametrize("value", ["     False ", "\\t\\t0\\t\\t", "  no\\t  "])
+def test_to_boolean_surrounding_whitespace(value):
+    """Test Surrounding whitespace"""
+    # Arrange
+
+    case_sensitive = True
+    # Act
+
+    actual = parse.to_boolean(
+        value, {"True", "1", "yes"}, {"False", "0", "no"}, case_sensitive=case_sensitive
+    )
+
+    # Assert
+    assert isinstance(actual, bool)
+    assert actual is False
+
+
+@pytest.mark.parametrize(
+    "value",
+    [
+        "I’m pretty confident this is true",
+        "null",
+        "",
+        "00",
+        "01",
+        "10",
+        "TrueFalse",
+        "True False",
+        "-",
+        "    ",
+        "Τʀυe",
+    ],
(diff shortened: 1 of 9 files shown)


mlipschultz:
Before this merges, I'd like to get approval on the tests (either here or on the [Confluence page](https://tetrascience.atlassian.net/wiki/spaces/DE/pages/3230400808/Unit+Test+Cases+for+Parsing+Primitives)) from @wkopen, @evan-anderson-ts, @igridchyn, @tlitetrasci, or @robjhornby-ts.  Basically, I want to make sure that someone very familiar with these parsing needs approves our test cases before we start researching/implementing a solution to pass those test cases, otherwise we might waste time/effort on the wrong solution.

task_script_utils/parse.py
```diff
@@ -0,0 +1,89 @@
...
+def to_boolean(
+    value: str, true_set: Set[str], false_set: Set[str], case_sensitive: bool = False
+) -> Optional[bool]:
+    """
+    Converts given string to a boolean value based on whether it is in either of the sets provided
+
+    :param value: A string to convert
+    :type value: string
+    :param true_set: A set of strings which represent True values
+    :type true_set: Set[string]
```
VladimirShteyn, line 64: IMO we shouldn't specify type in the docstring. Doing so creates two sources of truth: one in the function signature and another in the docstring. I recommend deleting `:type ...:` throughout.

__tests__/unit/parse_functions/test_to_int.py
```diff
@@ -0,0 +1,243 @@
...
+@pytest.mark.parametrize(
+    "value",
+    [
+        "two",
+        "12.4.5",
+        "1/2",
+        "4 + 2i",
+        "5 + 3j",
+        "2,500",
+        "0xbadbeef",
```
VladimirShteyn, line 231: 🥩

nom!

__tests__/unit/parse_functions/test_to_int.py
```diff
@@ -0,0 +1,243 @@
...
+    value = "-2.74e-3"
+
+    # Act
+    actual = parse.to_int(value)
+
+    # Assert
+    assert actual is None
+
+
+def test_to_int_leading_whitespace():
```
VladimirShteyn, line 146: [minor] For readability, consider grouping the whitespace tests into a `parametrize` or a single class.

__tests__/unit/parse_functions/test_to_int.py
```diff
@@ -0,0 +1,243 @@
...
+    """Test Scientific Notation, negative exponent, between 0 and 1."""
+    # Arrange from parameters
+    # Act
+    actual = parse.to_int(value)
+
+    # Assert
+    assert actual is None
+
+
+def test_to_int_scientific_notation_negative_exponent_greater_than_1():
```
VladimirShteyn, line 108: I recommend adding the following test cases where the exponent is zero,  probably in a `parametrize`. That is,

```
"1e0", "1e+0", "1e-0",
"1.0e0", "1.0e+0", "1.0e-0",
"100e0", "100e+0", "100e-0",
"100.0e0", "100.0e+0", "100.0e-0",
```

task_script_utils/parse.py
```diff
@@ -0,0 +1,89 @@
...
+    # --- Reference implementation ---
+    # try:
+    #     return Decimal(value)
+    # except Exception:
+    #     return None
+    raise NotImplementedError("Not implemented yet")
+
+
+def to_boolean(
+    value: str, true_set: Set[str], false_set: Set[str], case_sensitive: bool = False
```
VladimirShteyn, line 56: Should we have a default `true_set` and `false_set`? If so, I recommend

```
true_set: Set[str] = frozenset(('true', 'True', 1, 'yes', 'Yes'))
false_set: Set[str] = frozenset(('false', 'False', '0', 'no', 'No'))
```

__tests__/unit/parse_functions/test_to_decimal.py
```diff
@@ -0,0 +1,353 @@
...
+    """Test Negative decimal with trailing zeros."""
+    # Arrange
+    value = "-56.78000"
+
+    # Act
+    actual = parse.to_decimal(value)
+
+    # Assert
+    assert isinstance(actual, Decimal)
+    assert actual.compare_total(Decimal("-56.78000")) == 0
```
VladimirShteyn, line 122: TIL `compare_total`

VladimirShteyn (commented):
Nice work! I just had a few questions.

__tests__/unit/parse_functions/test_to_decimal.py
mlipschultz: Same here!

task_script_utils/parse.py
mlipschultz: For this PR, I think the answer's no.  I think it makes sense to have that discuss as part of this project though.

igridchyn (approved):
looks awesome! some minor comments

__tests__/unit/parse_functions/test_to_int.py
mlipschultz: Good idea!  Tests added

__tests__/unit/parse_functions/test_to_boolean.py
```diff
@@ -0,0 +1,202 @@
...
+        value, {"True", "1", "yes"}, {"False", "0", "no"}, case_sensitive=case_sensitive
+    )
+
+    # Assert
+    assert actual is False
+
+
+@pytest.mark.parametrize(
+    "value",
+    ["TRUE", "FALSE", "Yes", "yEs", "yeS", "YEs", "YeS", 'yES""YES', "No", "nO", "NO"],
```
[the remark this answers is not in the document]
mlipschultz: Good catch!

__tests__/unit/parse_functions/test_to_boolean.py
```diff
@@ -0,0 +1,202 @@
...
+    # Act
+    actual = parse.to_boolean(
+        value, {"True", "1", "yes"}, {"False", "0", "no"}, case_sensitive=case_sensitive
+    )
+
+    # Assert
+    assert actual is True
+
+
+@pytest.mark.parametrize("value", ["False", "0", "no"])
```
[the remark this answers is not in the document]
mlipschultz: For this test, we just want to know whether a value found in the `true_set` results in `True` being returned.  When it comes time to implement the `to_boolean` function, then we might want to have a comprehensive list.

__tests__/unit/parse_functions/test_to_decimal.py
```diff
@@ -0,0 +1,353 @@
...
+
+    # Act
+    actual = parse.to_decimal(value)
+
+    # Assert
+    assert isinstance(actual, Decimal)
+    assert actual.compare_total(Decimal("3.1415")) == 0
+
+
+def test_to_decimal_decimal_representation_of_integer():
```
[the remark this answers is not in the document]
mlipschultz: It's probably not needed, but it'd be embarrassing if somehow we ended up with that bug.  Test added

__tests__/unit/parse_functions/test_to_float.py
```diff
@@ -0,0 +1,325 @@
+"""Test `to_float()` function in parse.py."""
```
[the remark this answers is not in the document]
mlipschultz: I thought we talked about this when deciding on test cases, but I can't find it in the confluence page.  I guess we somehow missed this :disappointed:.  I'll update the confluence page.  Thanks!

__tests__/unit/parse_functions/test_to_boolean.py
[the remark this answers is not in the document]
abclay: The design so far is to not have any comprehensive list of false/true strings.  That is up to the developer to pass into the function.  We cannot assume that we know which values map to False/True in all cases, and the task script dev should know.  It makes the function more clear.

__tests__/unit/parse_functions/test_to_boolean.py
```diff
@@ -0,0 +1,185 @@
...
+    # Assert
+    assert actual is None
+
+
+@pytest.mark.parametrize(
+    "value",
+    [
+        "    True",
+        " 1",
+        "\\t   \\t yes",
```
abclay, line 135: This looks like my fault, but I think these `\\t` values should be `\t`

__tests__/unit/parse_functions/test_to_float.py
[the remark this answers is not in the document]
mlipschultz: Discussion had and decisions made.  Thanks again @igridchyn!

__tests__/unit/parse_functions/test_to_boolean.py
```diff
@@ -109,7 +109,19 @@ def test_to_boolean_case_insensitive_string_is_in_false_set(value):
...
-    ["TRUE", "FALSE", "Yes", "yEs", "yeS", "YEs", "YeS", 'yES""YES', "No", "nO", "NO"],
+    [
+        "TRUE",
+        "FALSE",
+        "Yes",
+        "yEs",
+        "yeS",
+        "YEs",
+        "YeS",
+        'yES", "YES',
```
mlipschultz, line 120: black failed me :angry:

__tests__/unit/parse_functions/test_to_boolean.py
mlipschultz: Fixed

abclay (commented):
I can't approve because I'm the original author, but I approve of these changes anyways.

__tests__/unit/parse_functions/test_to_decimal.py
```diff
@@ -239,93 +239,185 @@ def test_to_decimal_scientific_notation_negative_exponent_greater_than_1():
...
-    value = "128.28       "
+    value = "-512.00"
 
     # Act
     actual = parse.to_decimal(value)
 
     # Assert
     assert isinstance(actual, Decimal)
-    assert actual.compare_total(Decimal("128.28")) == 0
+    assert actual.compare_total(Decimal("-512.00")) == 0
```
VladimirShteyn, line 276: [minor] According to [the compare_total docs](https://docs.python.org/3/library/decimal.html#decimal.Decimal.compare_total) this should be `assert actual.compare_total(Decimal("-512.00")) == Decimal("0")

__tests__/unit/parse_functions/test_to_decimal.py
```diff
@@ -341,6 +433,10 @@ def test_to_decimal_string_has_underscores_decimal():
         "0xbadbeef",
         "0o775",
         "0b100001",
+        "infinity",
+        "INFINITY",
```
VladimirShteyn, line 437: ALL CAPS for when you need a higher cardinality infinity.

VladimirShteyn (approved):
LGTM. To make the code more readable, might be worth breaking up long strings into multiple lines as suggested [in this reviewer comment](https://github.com/tetrascience/ts-task-script-utils/pull/21/files/35862c5276c7320d42bdd7bd536b4707d86c665d..53675468c0de2c54814503b3d55c682bfa5e98fa#r857863373).

__tests__/unit/parse_functions/test_to_decimal.py
```diff
@@ -0,0 +1,407 @@
...
+
+    # Act
+    actual = parse.to_decimal(value)
+
+    # Assert
+    assert (
+        actual.compare_total(
+            Decimal(
+                "114813069527425452423283320117768198402231770208869520047764273682576626139237031385665948631650626991844596463898746277344711896086305533142593135616665318539129989145312280000688779148240044871428926990063486244781615463646388363947317026040466353970904996558162398808944629605623311649536164221970332681344168908984458505602379484807914058900934776500429002716706625830522008132236281291761267883317206598995396418127021779858404042159853183251540889433902091920554957783589672039160081957216630582755380425583726015528348786419432054508915275783882625175435528800822842770817965453762184851149029376.0"
```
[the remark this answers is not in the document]
mlipschultz: I was hoping Black would do this for me, but it didn't :angry:

task_script_utils/parse.py
```diff
@@ -19,21 +19,6 @@ def to_int(value: str) -> Optional[int]:
     raise NotImplementedError("Not implemented yet")
 
 
-def to_float(value: str) -> Optional[float]:
```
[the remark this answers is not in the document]
mlipschultz: Should we create a `to_float` function that just calls `to_decimal` and convert the output?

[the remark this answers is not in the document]
abclay: Would that be part of this effort?  Or would we put the unit tests back and strongly suggest the implementer decide to use `to_decimal` as the parser within `to_float`?

__tests__/unit/parse_functions/test_to_decimal.py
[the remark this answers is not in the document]
abclay: for really long ints, we could do something like 

```python
expected = int(
    "123456790"
    "123456790"
    "123456790"
    "123456790"
)
assert expected == actual
```
with the understanding that we trust the `int()` built-in.  Or we could just leave them as-is and mark these files telling pylint to ignore long lines in them.

task_script_utils/parse.py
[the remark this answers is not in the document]
abclay: +1 to bringing back `to_float` for right now.  We can base its implementation on `to_decimal` and deprecate it later if we want.  Folks should be able to switch over to `to_decimal` easily since it's the same implementation and behaves the same except for the output type.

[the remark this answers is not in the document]
mlipschultz: Brought back and added tests for nan and inf

abclay (commented):
LGTM

PR #25, opened 2022-04-27
DE-3695: Prepare for v1.2.0 release

abclay:
## JIRA Ticket

https://tetrascience.atlassian.net/browse/DE-3695

- [x] PR Title should be in the form: "DE-12345: Short Description of Changes"

## Description of Changes

* Updated version number in `setup.py`
* Ran `pipenv upgrade` to get latest versions in lock file
* Ran `pipenv lock -r > requirements.txt` to get latest requirements for publishing

## General

- [x] Not applicable
- [ ] Changelog is added in README.md

### Testing Completed

- [x] Tested manually locally
- [ ] Unit tests updated
- [ ] None (please explain)

## Related Information

<!--
Please list any related information here.
For example: links to other PRs, link to IDS, link to confluence docs, etc
--!>


VladimirShteyn (changes requested):
We're pinning dependency versions in the setup.py file, which is a big no-no. [link to code](https://github.com/tetrascience/ts-task-script-utils/pull/25/files#diff-60f61ab7a8d1910d86d9fda2261620314edcae5894d5aaa236b821c7256badd7R12-R15). can we fix that in this PR? It's caused dependency conflict issues for DI Team members. It would be unfortunate if we propagated this issue to v1.2.0.

abclay:
> We're pinning dependency versions in the setup.py file, which is a big no-no. [link to code](https://github.com/tetrascience/ts-task-script-utils/pull/25/files#diff-60f61ab7a8d1910d86d9fda2261620314edcae5894d5aaa236b821c7256badd7R12-R15). can we fix that in this PR? It's caused dependency conflict issues for DI Team members. It would be unfortunate if we propagated this issue to v1.2.0.

Makes sense to me.  I'll look into how to fix.

abclay:
I switched the dependency management from pipenv to poetry.  `poetry publish` is so easy!

I also fixed the `black` issue I introduced (wrong place to make the line length long...) and removed the Travis integration.

pyproject.toml
```diff
@@ -1,12 +1,39 @@
...
+
+[tool.poetry.dependencies]
+# python version is restricted to less than 3.11 to allow numpy to resolve correctly
+python = "^3.7,<3.11"
+python-dateutil = "^2.8.2"
+dateparser = "^1.1.1"
+arrow = "^1.2.2"
+pendulum = "^2.1.2"
+pydash = "^5.1.0"
+numpy = "^1.21.0"
```
VladimirShteyn, line 30: [minor] We could loosen this to `>=1.17.0,<2`. I don't see any features in `ts-task-script-utils` that require a version higher than `1.17.0`.

VladimirShteyn (approved):
Good work. Very clean.

pyproject.toml
VladimirShteyn: I just noticed that one of the dependencies is `pytest-cov`. It causes issues running individual tests. Perhaps we should include the following text in README.md or CONTRIBUTING.md?

```markdown
## Running tests

`ts-task-script-utils` uses `pytest` as its testing framework. As such, tests can be executed with the `pytest` shell command (see the [pytest docs here](https://docs.pytest.org/en/latest/how-to/usage.html)). Note that `pytest-cov`, a `pytest` plugin dependency of `ts-task-script-utils`, prevents one from running individual tests; any attempt to run a single test runs the entire test module. To turn off `pytest-cov` and thus enable running individual tests, pass the `--no-cov` flag to the `pytest` command. For example, to run the `TestMyExampleClass.test_my_examle_method` test located at `tests/test_my_module.py`, do

```shell
pytest --no-cov tests/test_my_module.py::TestMyExampleClass::test_my_examle_method
```
```

abclay: I don't think that running tests belongs in the README, since that goes straight onto PyPI.  I think creating a CONTRIBUTING file is outside the scope of this effort, but I do think it's something worth doing.  Would you mind creating a ticket for that?

mlipschultz (approved):

pyproject.toml
VladimirShteyn: Can do.

PR #28, opened 2022-05-19
DE-3885: Implement parsing functions

VladimirShteyn:
## JIRA Ticket
https://tetrascience.atlassian.net/browse/DE-3885

## Description of Changes
Updated the implementation of the parsing functions, mostly by deleting the research implementations. The only major change in logic was improving the error message of `to_boolean`.

## General

- [ ] Not applicable
- [ ] Changelog is added in README.md

### Testing Completed

- [x] Tested manually locally
- [x] Unit tests updated
- [ ] None (please explain)


task_script_utils/parse.py
```diff
@@ -190,17 +80,51 @@ def to_boolean(
     :raises ValueError: If the sets overlap or if either set is empty
     """
     if not true_set or not false_set:
-        raise Exception("Fail")
-    if len(true_set & false_set):
-        raise Exception("Fail")
-    true_set_compare = true_set
-    false_set_compare = false_set
+        true_set_quoted_items = [f"'{true_value}'" for true_value in true_set]
```
mlipschultz, line 83: What's the reason for quoting the values for this error?  Could the message for `ValueError` just be:
```python
        raise ValueError(
            f"The set of 'true' values and 'false' values cannot be empty. "
            f"true_set: {true_set}. "
            f"false_set: {false_set}"
        )
```

__tests__/unit/parse_functions/test_to_boolean.py
```diff
@@ -3,30 +3,51 @@
...
-    """Test Empty True Set."""
-    # Arrange
-
-    # Act
-
-    # Assert
-    with pytest.raises(Exception):
-        parse.to_boolean(value, set(), {"False"}, case_sensitive=case_sensitive)
+class TestToBooleanErrors:
+    """Tests for cases where `to_boolean` raises an exception."""
```
abclay, line 10: Is this to make testing in your IDE easier?  I understand the desire to group tests, but I'm not a big fan of having to add `@staticmethod` for all the test functions.

pyproject.toml
```diff
@@ -28,9 +28,7 @@ arrow = "^1.2.2"
 pendulum = "^2.1.2"
 pydash = "^5.1.0"
 numpy = "^1.17.0"
-fastnumbers = "^3.2.1"
 pandas = "1.3.5"
```
abclay, line 31: Are numpy and pandas still dependencies in this library?  I don't remember.

task_script_utils/parse.py
```diff
@@ -190,17 +80,51 @@ def to_boolean(
...
+            f"false_set: ({', '.join(false_set_quoted_items)})"
+        )
+
     value_compare = value.strip()
     if not case_sensitive:
-        value_compare = value_compare.lower()
-        true_set_compare = {t.lower() for t in true_set}
-        false_set_compare = {f.lower() for f in false_set}
+        def to_lowercase(string):
+            return string.casefold()
```
mlipschultz, line 94: :trophy:

abclay (commented):
Couple quick questions from me.  I will hold off on approving because I wrote the original code and someone else should sign off.

task_script_utils/parse.py
```diff
@@ -190,17 +80,51 @@ def to_boolean(
...
+            false_intersection.update(false_set_compare[value])
+
+        raise ValueError(
+            f"'true_set' and 'false_set' have values in common. "
+            f"true_set: ({', '.join(true_intersection)}). "
+            f"false_set: ({', '.join(false_intersection)})."
+        )
 
+    true_set_compare: Set[str] = set(true_set_compare)
+    false_set_compare: Set[str] = set(false_set_compare)
```
mlipschultz, line 127: Why convert the keys into a set before doing the contains check?

mlipschultz (changes requested):
There's one or two deviations from our agreed-upon plan for v1.  I think those should be fixed.  I also have a couple of other small improvements suggested.

pyproject.toml
mlipschultz: Numpy is [currently a dependency](https://github.com/tetrascience/ts-task-script-utils/blob/master/pyproject.toml#L30), but not pandas.

.gitignore
```diff
@@ -13,3 +13,4 @@ dev.*
 .idea/
 venv/
 __pycache__/
+**/.python-version
```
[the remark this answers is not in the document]
VladimirShteyn: `.python-version` is a file used by `pyenv` to specify the folder's version of Python.

pyproject.toml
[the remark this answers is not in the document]
VladimirShteyn: Done.

task_script_utils/parse.py
```diff
@@ -1,160 +1,50 @@
...
+    if "_" in value:
+        return None
+    if "e" in value.lower() or "." in value:
         try:
-            array = np.asarray([value], dtype=float)
-            int_val = int(array[0])
-            if int_val == array[0]:
+            float_val = float(value)
+            int_val = int(float_val)
+            if np.isclose(int_val, float_val):
```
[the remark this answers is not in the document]
VladimirShteyn: In some cases, this implementation is consistent with our earlier decision and in others it's not. Could `float("1")` be evaluated to `1.000000000000001` on some systems, thus erroneously returning `None`? That said, I agree, `None` is the agreed return value when the user passes `to_int("1.000000000000001")`.

task_script_utils/parse.py
VladimirShteyn: Yeah, let's keep it simple. Sometimes it's helpful to visually separate values with quotes, but the commas already do that.

task_script_utils/parse.py
VladimirShteyn: Good point!

__tests__/unit/parse_functions/test_to_boolean.py
VladimirShteyn: Yes, and to make the test names shorter and thus easier to read. I've configured PyCharm to use a test template, so adding `@staticmethod` is no extra work. One can also use a vanilla instance method for tests too, but that's an extra `self` to type. Not sure which is worse.

task_script_utils/parse.py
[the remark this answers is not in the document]
VladimirShteyn: I changed `math.isclose(int_val, float_val)` to `int_val == float_val`.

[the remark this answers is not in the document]
mlipschultz: Did you push that change yet?  I'm still seeing `math.isclose`.

I agree though that we might want to discuss what "close enough" is in a future iteration.

mlipschultz (approved):

pyproject.toml
VladimirShteyn: I found I could remove `numpy` as a dependency and the tests still pass. 🎉

PR #31, opened 2022-06-23
DE-4184: make parse the only entrypoint

setiat:
## JIRA Ticket

https://tetrascience.atlassian.net/browse/DE-4184
- [x] PR Title should be in the form: "DE-12345: Short Description of Changes"

## Description of Changes

- deprecate `convert_datetime_to_ts_format`
- undepricate `parse`
- removed `_parse_with_formats`
- updated test cases
- updated readmes to use `parse` instead of `parse_with_fromats`

## General

- [ ] Not applicable
- [x] Changelog is added in README.md

### Testing Completed

- [ ] Tested manually locally
- [x] Unit tests updated
- [ ] None (please explain)

## Automated testing

- [ ] Unit tests
- [ ] Integration tests
- [ ] API tests
- [ ] End-To-End tests
- [ ] N/A (please clarify)


setiat:
@abclay, @apjansing , @mlipschultz 
I would also like to remove changelogs from datetime_parser/readme.md. Feels redundant. instead use task_script_utils/readme.md
What do you think?

task_script_utils/datetime_parser/README.md
```diff
@@ -451,6 +452,11 @@ A `DatetimeConfig` object has following attributes:
 
 ## Changelog
 
+### v1.3.1
+
+- Remove`utils._parse_with_formats`. `utils.parse_with_formats` will be used instead.
+- Un-depricate `parse`
```
mlipschultz, line 458:
```suggestion
- Un-deprecate `parse`
- Mark `convert_datetime_to_ts_format.convert_datetime_to_ts_format` as deprecated
```

task_script_utils/datetime_parser/utils/parsing.py
```diff
@@ -21,49 +21,16 @@
...
-        config (DatetimeConfig, optional): Datetime Configuration.
-        Defaults to DEFAULT_DATETIME_CONFIG.
-
-    Raises:
-        DatetimeParserError: When datetime_str can be parsed into TSDatetime object
-    Returns:
-        TSDatetime
-    """
-    parsed_datetime = None
+) -> Tuple[Optional[TSDatetime], Optional[str]]:
```
mlipschultz, line 27: This is a change to the API.  Any task script already using the old `parse_with_formats` will break when it starts using this change.  I'm not saying the change is wrong, just that it's something we need to think about.

Because it's a change to the API, semantic versioning says that's a bump to the major number, but I see you bumped the patch.

Why change the return type of `parse_with_formats` though?  My understanding was that it's unclear what was deprecated.

mlipschultz (changes requested):
One comment is a fairly small change to the changelog.  The other is raising a question for discussion about a change to the API -- before making any changes, it'd be good for us to decide what the correct action is.

abclay:
> @abclay, @apjansing , @mlipschultz I would also like to remove changelogs from datetime_parser/readme.md. Feels redundant. instead use task_script_utils/readme.md What do you think?

That makes sense to me.

task_script_utils/datetime_parser/utils/parsing.py
mlipschultz: > I will check if any task-script is using

Please do.  Was `parse_with_format` mentioned in the documentation?  If not, then it could be safe to assume no one's using it.

If `parse_with_format` is meant to be internal-only then I'm fine bumping the patch number.

